PR #31, opened 2026-03-31
Update dashboard sessions to use agent data and show upcoming tasks

bradbrok:
## Summary
- replace the dashboard's legacy session list with agent-backed streaming and session health data
- make active sessions show meaningful agent, state, work, context, and cost details
- remove the outreach platforms dashboard section and replace its summary stat with open tasks
- add an upcoming tasks table with priority, status, owner, and due-date labeling
- include the rebuilt frontend assets in `frontend-dist`

## Testing
- `npm run build`


olegbrok (approved):
LGTM — solid upgrade to the dashboard.

**What's good:**
- Replacing the static outreach platforms table with an upcoming tasks view is the right call — way more actionable
- Session normalization logic (streaming vs legacy) is clean and handles the merge correctly with seenIds dedup
- Task sorting by due date → priority → created_at is sensible
- The expanded session detail now shows operationally useful info (task counts, reconnects, errors) instead of raw token counts
- `formatDueLabel` with relative day labels is a nice touch

**Minor notes (non-blocking):**
- `loadSessions` fires N×3 parallel API calls (streaming-sessions, sessions, health per agent). If the agent count grows, this could get chatty. Not a problem today, but worth keeping in mind
- The `refresh()` function awaits `loadSessions` inside the main try block — if one agent's endpoint hangs, the whole dashboard stalls. The per-agent `.catch(() => ...)` helps but the outer `await` is still serial after the initial batch
- Built assets in the PR are fine since they're in `frontend-dist`

Ship it.

PR #146, opened 2026-04-07
Fix chat routing for named streaming sub-sessions

bradbrok:
## Summary
- route chat history, polling, uploads, restart/archive/compact, and model actions to the selected streaming session instead of always targeting `main`
- give streaming sessions stable `{agent}-{label}` ids and expose them from the API so the UI can treat sub-sessions as first-class chats
- prompt for a session name when creating a new chat session and normalize it into a safe label before opening it

## Testing
- `pytest tests/test_api.py -q`
- `npm run build`

### DIFF
--- a/frontend-svelte/src/pages/Chat.svelte
+++ b/frontend-svelte/src/pages/Chat.svelte
@@ -47,6 +47,8 @@
     let sessionsList = [];
     let activeSession = null;
     let activeAgent = null;
+    let activeSessionLabel = 'main';
+    let activeSessionStreaming = false;
     let messages = [];
     let messageInput = '';
     let sending = false;
@@ -118,7 +120,30 @@
                 if (!matched) orphans.push(s);
             }
         }
+        for (const sessionGroup of Object.values(groups)) {
+            sessionGroup.sort((a, b) => {
+                const aMain = (a.label || '') === 'main' ? 1 : 0;
+                const bMain = (b.label || '') === 'main' ? 1 : 0;
+                if (aMain !== bMain) return bMain - aMain;
+                return (b.last_active || 0) - (a.last_active || 0);
+            });
+        }
         return { groups, orphans };
+    }
+
+    function inferAgentName(sessionId) {
+        if (!sessionId || !sessionId.includes('-')) return '';
+        return sessionId.split('-')[0];
+    }
+
+    function normalizeSessionLabel(raw) {
+        return (raw || '')
+            .trim()
+            .toLowerCase()
+            .replace(/\s+/g, '-')
+            .replace(/[^a-z0-9_-]/g, '')
+            .replace(/-+/g, '-')
+            .replace(/^[-_]+|[-_]+$/g, '');
     }
 
     async function refreshSessions() {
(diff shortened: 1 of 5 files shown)


olegbrok (commented):
## Barsik's Review — PR #146

**Overall:** Solid backend work on the sub-session routing problem. The backend/broker/streaming changes are clean and well-tested. However, the frontend changes will have **massive merge conflicts** with `main` since we already refactored Chat.svelte into components (ChatSidebar, ChatMessage, ChatInput) + extracted chatUtils.js.

### 👍 What's good (keep these)

- **`_require_streaming_session` helper** — DRY pattern, eliminates repeated lookup + 404 logic across 5+ endpoints. Clean.
- **Consistent `?label=` param** — all streaming endpoints now accept label, not just some. Good API design.
- **`StreamingSession.id` property** — returning `f"{agent_name}-{label}"` is the right abstraction.
- **`StreamingSessionConfig.label` field** — explicit label in config, no more guessing from session_id parsing.
- **Broker `_get_streaming_session` label kwarg** — direct lookup instead of always falling back to main.
- **4 new tests** — `test_chat_label_routes_to_subsession`, `test_streaming_status_label`, `test_require_streaming_session_missing`, `test_subsession_chat_end_to_end`. Good coverage.

### ⚠️ Frontend conflicts

The PR's Chat.svelte is based on the **pre-refactor monolithic version** (~1800 lines). On `main`, we already:
- Extracted `ChatSidebar.svelte` (agent list, sessions, rename)
- Extracted `ChatMessage.svelte` (all message rendering variants)
- Extracted `ChatInput.svelte` (input bar, file upload, reply)
- Extracted `chatUtils.js` (parseBrokerMessage, groupByAgent, sortMessages, etc.)
- Added session rename with conversation history migration
- Added reply-to, forward modal, session events, heartbeat filtering

Merging this PR's frontend changes as-is would **overwrite all of that** and regress those features.

### 🔧 Recommendation

**Cherry-pick the backend changes, skip the frontend rewrite:**

1. ✅ Take: `api.py` changes (`_require_streaming_session`, `?label=` params, rename endpoint)
2. ✅ Take: `broker.py` label kwarg
3. ✅ Take: `streaming_session.py` id property + config label
4. ✅ Take: All 4 new tests
5. ❌ Skip: `Chat.svelte` — apply the routing logic (`normalizeSessionLabel`, `prompt()` naming) on top of the already-refactored component structure instead

The frontend concepts (label normalization, prompt for session name) are good ideas — they just need to be applied to the current component structure, not the old monolith.

### Minor notes
- `normalizeSessionLabel()` is a nice util — should go in `chatUtils.js` when ported
- The `prompt()` dialog for naming sessions could be a proper modal component for consistency with the rest of the UI

🤖 Reviewed by Barsik

olegbrok (commented):
## Barsik's Review — PR #146

**Overall:** Solid backend work on the sub-session routing problem. The backend/broker/streaming changes are clean and well-tested. However, the frontend changes will have **massive merge conflicts** with `main` since we already refactored Chat.svelte into components (ChatSidebar, ChatMessage, ChatInput) + extracted chatUtils.js.

### 👍 What's good (keep these)

- **`_require_streaming_session` helper** — DRY pattern, eliminates repeated lookup + 404 logic across 5+ endpoints. Clean.
- **Consistent `?label=` param** — all streaming endpoints now accept label, not just some. Good API design.
- **`StreamingSession.id` property** — returning `f"{agent_name}-{label}"` is the right abstraction.
- **`StreamingSessionConfig.label` field** — explicit label in config, no more guessing from session_id parsing.
- **Broker `_get_streaming_session` label kwarg** — direct lookup instead of always falling back to main.
- **4 new tests** — good coverage of the routing changes.

### ⚠️ Frontend conflicts

The PR's Chat.svelte is based on the **pre-refactor monolithic version** (~1800 lines). On main, we already:
- Extracted ChatSidebar.svelte, ChatMessage.svelte, ChatInput.svelte
- Extracted chatUtils.js (parseBrokerMessage, groupByAgent, sortMessages, etc.)
- Added session rename with conversation history migration
- Added reply-to, forward modal, session events, heartbeat filtering

Merging this PR's frontend changes as-is would **overwrite all of that** and regress those features.

### 🔧 Recommendation

**Cherry-pick the backend changes, skip the frontend rewrite:**

1. ✅ Take: api.py changes (_require_streaming_session, ?label= params)
2. ✅ Take: broker.py label kwarg
3. ✅ Take: streaming_session.py id property + config label
4. ✅ Take: All 4 new tests
5. ❌ Skip: Chat.svelte — apply the routing logic on top of the already-refactored component structure instead

The frontend concepts (normalizeSessionLabel, prompt for session name) are good ideas — they just need to be applied to the current component structure, not the old monolith.

🤖 Reviewed by Barsik

olegbrok (commented):
## Code Review — PR #146

**tl;dr:** Backend changes are correct and well-structured — safe to cherry-pick. Frontend diff is based on the old monolithic `Chat.svelte` and can't land as-is since `main` already has the component extraction (`ChatSidebar`, `ChatMessage`, `ChatInput`, `chatUtils.js`). One naming inconsistency to fix before cherry-picking.

---

### ✅ `streaming_session.py` — Core fix, correct

```python
# Before
@property
def id(self) -> str:
    return f"{self.agent_name}-main"  # always wrong for sub-sessions

# After
return f"{self.agent_name}-{self.label}"
```

This is the root fix everything else depends on. Adding `label` to `StreamingSessionConfig` and propagating it to `self.label` is clean. The `label or "main"` default in `__init__` handles empty strings gracefully.

---

### ✅ `broker.py` — Clean, backward compatible

The `label` keyword-only param shortcut in `_get_streaming_session` is correct:
```python
if label:
    return sessions.get(label)
# falls through to existing chat_id routing
```

Adding `"id": s.id` to `list_streaming_sessions` is the right call — the frontend was already having to reconstruct this.

---

### ✅ `api.py` — `_require_streaming_session` helper is the right pattern

```python
def _require_streaming_session(name: str, label: str = "main"):
    resolved_label = label or "main"
    ss = broker._get_streaming_session(name, label=resolved_label)
    if not ss:
        raise HTTPException(404, f"No streaming session '{resolved_label}' for '{name}'")
    return ss, resolved_label
```

Used consistently across `restart`, `model`, making them all label-aware. DRY improvement over the pattern in `main` that inlines this check each time. The `resolved_label` return value cleanly handles `label=""` edge cases.

`_session_id_for_agent(agent_name, label)` generalization is backward-compatible (defaults to `"main"`).

---

### ✅ Tests — 4 new tests, all substantive

- `test_streaming_session_uses_label_specific_conversation_id` — directly verifies the core invariant: worker and main sessions write to separate conversation IDs. Exactly the right test.
- `test_chat_endpoint_routes_to_selected_streaming_label` — verifies routing isolation (worker receives, main doesn't).
- `test_streaming_status_supports_selected_label_and_main_default` — verifies both explicit and default label paths return correct `id`/`label`.
- `test_streaming_restart_uses_selected_label` — verifies restart isolation (worker restarts, main doesn't disconnect).

Good coverage. These would have caught the original bug.

---

### ⚠️ Naming inconsistency: `?session=` vs `?label=`

`main` currently has the chat endpoint as:
```python
async def chat_with_agent(name: str, req: dict, session: str = "main"):
```

This PR changes it to `label: str = "main"`. `label` is the right name — it matches every other endpoint in the file. But cherry-picking the backend means this needs a decision: either update `main`'s chat endpoint to rename `session` → `label` at the same time, or this PR needs to keep `session` for compatibility. I'd recommend renaming to `label` for consistency, and updating the one reference in `Chat.svelte` (or `ChatInput`/Chat parent) that currently sends `?session=`.

---

### ❌ Frontend diff — can't land as-is

`main` already has:
- `Chat.svelte` refactored into `ChatSidebar`, `ChatMessage`, `ChatInput` components
- `chatUtils.js` for pure utilities
- Session enrichment logic from `b3ea3c0`

This PR's frontend is against the old monolith. Applying it would revert the component extraction.

The backend-only logic in the frontend diff is worth porting manually:
- `normalizeSessionLabel()` — good input sanitization, worth adding to `chatUtils.js`
- `activeSessionLabel` / `activeSessionStreaming` state vars — the component extraction already handles this differently via `_streaming_label` / `_from_store` flags, but the explicit state approach here is cleaner and easier to reason about
- `prompt()` / `alert()` for session naming — works but `prompt()`/`alert()` are modal-blocking browser dialogs; the existing auto-label approach (`chat-${Date.now().toString(36)}`) with an inline rename UX is better

---

### Recommendation

**Cherry-pick:** `streaming_session.py`, `broker.py`, `api.py`, `tests/test_api.py` — all safe to land on `main` after renaming `?session=` → `?label=` on the chat endpoint.

**Skip:** `frontend-svelte/src/pages/Chat.svelte` — would revert the component extraction. Port `normalizeSessionLabel` manually if wanted.

PR #181, opened 2026-04-10
Centralize onboarding CTA and clarify wizard flow

bradbrok:
### Motivation
- The onboarding wizard had duplicated/conflicting primary actions (in-step buttons + footer Next) that made progression confusing and error-prone.  
- The goal is to make the primary CTA deterministic per step, prevent accidental progression without required inputs, and make the flow clearer for fresh and partially-configured installs.

### Description
- Added derived state variables `hasAnyAgent` and `hasConfiguredChannel` to reflect whether an agent or a channel is already present.  
- Introduced a centralized `handlePrimaryAction()` that performs the correct action per step (save profile, create agent, configure channel, complete onboarding, or advance).  
- Removed duplicated in-step primary buttons for Profile and Agent steps and replaced them with guidance copy pointing to the unified bottom-right CTA.  
- Updated the footer CTA to use `handlePrimaryAction`, with conditional disabling and dynamic labels so the primary button performs appropriate actions and prevents invalid progression.  
- Modified file: `frontend-svelte/src/pages/Onboarding.svelte`.

### Testing
- Ran the frontend production build with `cd frontend-svelte && npm run build`, which completed successfully.  
- The build emitted non-blocking Vite/Svelte warnings about chunk size and an SSR button-in-button placement warning, but these were not blocking and the build finished OK.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d84a6000dc8333b306be0081228c56)

### DIFF
--- a/frontend-svelte/src/pages/Onboarding.svelte
+++ b/frontend-svelte/src/pages/Onboarding.svelte
@@ -55,6 +55,8 @@
     let platformTestResult = null;
     let platformTesting = false;
     let configuredPlatforms = [];
+    $: hasAnyAgent = agentCreated || existingAgents.length > 0;
+    $: hasConfiguredChannel = platformConfigured || configuredPlatforms.length > 0;
 
     // Load data for current step
     async function loadStepData() {
@@ -305,6 +307,19 @@
     function next() {
         if (step < totalSteps - 1) { step++; loadStepData(); }
     }
+    async function handlePrimaryAction() {
+        if (step === totalSteps - 1) return completeOnboarding();
+        if (step === 3) return saveProfile(true);
+        if (step === 4) {
+            if (hasAnyAgent) return next();
+            return createAgent();
+        }
+        if (step === 5) {
+            if (hasConfiguredChannel) return next();
+            return configurePlatform();
+        }
+        return next();
+    }
 
     function keyLabel(k) {
         return k.replace('_API_KEY', '').replace(/_/g, ' ');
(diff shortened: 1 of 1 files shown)


olegbrok (changes requested):
## Combined Review — Barsik + Pushok

### What's good
- `handlePrimaryAction()` dispatch logic is clean and correct
- `hasAnyAgent` / `hasConfiguredChannel` derived state is a nice DRY pattern
- The *idea* of centralizing the footer CTA is sound

### Request Changes

**1. UX Regression — Removing in-step buttons is a step backward**
We just fixed the onboarding flow because users couldn't find the Next button and expected the contextual action buttons to advance. This PR removes those in-step buttons and replaces them with passive hint text ("Save and continue from the bottom right when you are ready"). That re-introduces the exact problem we solved.

The user's eye is on the form. The action should be right there, not in a footer they have to hunt for — especially on mobile where the footer may be below the fold.

**2. Step 5 (Channel): Configure button removed from step body**
The `{:else}` block that showed the configure button when `!platformConfigured` is deleted. Now the only way to configure is the footer CTA, but the Test button is still inline. You can test a connection (inline) but not configure one (footer only). Inconsistent and confusing.

**3. Step 3 (Profile): `ownerName` silently became required**
The footer button is now disabled when `!ownerName.trim()` on step 3. Previously profile was fully optional — you could skip through. This is a behavioral change buried in a refactor. If name should be required, that's a separate deliberate decision.

**4. Ternary monster on the footer button**
The inline disabled condition and label expression are hard to read and maintain:
```svelte
disabled={loading || (step === 3 && !ownerName.trim()) || (step === 4 && !hasAnyAgent && !agentDisplayName.trim()) || ...}
```
Should at minimum be extracted to `$: primaryDisabled = ...` and `$: primaryLabel = ...` computed vars.

**5. Hardcoded English strings**
The hint texts "Save and continue from the bottom right when you are ready" and "Create and continue from the bottom right when you are ready" aren't using `$_()` i18n keys.

### Suggested approach
Keep the contextual in-step buttons AND have the footer CTA call `handlePrimaryAction()`. Best of both worlds — in-step button is the primary path, footer is a consistent fallback. The `handlePrimaryAction` dispatch is still useful, just don't remove the inline buttons to get there.

🤖 Reviewed by Barsik + Pushok

PR #189, opened 2026-04-10
Consolidate pinky-memory into shared SSE server

### LINKED ISSUE #188: Consolidate pinky-memory from stdio to shared SSE server
## Summary

All three core MCP servers (pinky-self, pinky-messaging, pinky-memory) should run as shared SSE on the same server (port 8890). Currently pinky-memory is the last holdout — it spawns a separate stdio subprocess **per agent**, meaning ~15 Python processes just for memory at steady state.

The shared MCP code already flags this as planned: `shared_mcp.py:263` has a comment: *"pinky-memory is NOT included yet — it needs a store object per-agent, which requires a different approach. Phase 3."*

## Current State

| Server | Transport | Process Model |
|--------|-----------|---------------|
| pinky-self | SSE (shared) | 1 process, multi-agent via `X-Agent-Name` header |
| pinky-messaging | SSE (shared) | 1 process, multi-agent via `X-Agent-Name` header |
| **pinky-memory** | **stdio** | **1 subprocess per agent** (~15 processes) |
| pinky-web | stdio | per-agent, optional (skill-added) |
| pinky-calendar | stdio | per-agent, optional |

### Core Changes

1. **`pinky_memory/server.py`** — Refactor `create_server()` to accept a `store_pool: dict[str, ReflectionStore]` instead of a single `store`. Each tool resolves the current agent from `ContextVar` via `LazyAgentName`, then looks up `store_pool[agent_name]`.

2. **`pinky_daemon/shared_mcp.py`** — In `SharedMcpManager._create_app()`:
   - Build a store pool: one `ReflectionStore(db_path)` per registered agent
   - Share one `EmbeddingClient` across all agents (it's stateless)
   - Mount memory server at `/mcp/memory/sse`
   - Add `register_agent(name, db_path)` / `unregister_agent(name)` for runtime agent lifecycle

3. **`pinky_daemon/api.py` `_write_mcp_json()`** — Add memory SSE branch inside `if SHARED_MCP_ENABLED:`, pointing to `http://127.0.0.1:8890/mcp/memory/sse` with `X-Agent-Name` header.

4. **No migration needed** — `.mcp.json` files are regenerated at daemon startup automatically.

### Already Working in Our Favor

- `ReflectionStore` is thread-safe (uses `threading.RLock`, `check_same_thread=False`)
- `create_server()` is already decoupled from DB path — takes a `store` object
- `LazyAgentName` pattern already solves per-request agent routing
- Standalone SSE mode (`python -m pinky_memory --transport sse`) already exists
- Daemon's `_get_memory_store()` already opens stores on-demand for REST endpoints — same pattern

## Benefits

- **~15 fewer processes** at steady state (one per agent eliminated)
- **Faster agent startup** — no subprocess spawn, store already in pool
- **Single embedder instance** instead of 15 idle OpenAI clients
- **Consistent management** — all core MCPs under one supervisor with crash recovery
- **Simpler `.mcp.json`** — core servers all become SSE URLs

## Risks / Watch-outs

- **No auth on agent routing** — `X-Agent-Name` header is trusted, no crypto verification. Same risk already accepted for pinky-self/messaging. Localhost-only, acceptable for now.
- **Store pool lifecycle** — must handle agent creation/deletion at runtime, lazy init for agents with no memory.db yet, and graceful error for unknown agent names.
- **Embedder is synchronous** — blocks shared MCP thread pool during embedding calls. Already accepted tradeoff. No change needed.
- **Connection count** — 15 persistent SQLite connections is fine; at 100+ agents would want lazy eviction. Not a concern now.

## Also Consolidatable (Lower Priority)

- **pinky-web** — stateless, no per-agent routing needed. Only used by barsik currently.
- **pinky-calendar** — per-agent OAuth tokens, same pool pattern as memory. Optional, not widely used.

🤖 Opened by Barsik

olegbrok:
## Summary

Closes #188. Moves pinky-memory from per-agent stdio subprocesses to the shared SSE server (port 8890), completing the MCP consolidation. All three core MCP servers now run in a single process.

- **`pinky_memory/server.py`**: `create_server()` now accepts an optional `store_factory` callable. In shared SSE mode, each tool resolves the agent from `ContextVar` and gets the right `ReflectionStore` via the factory. Stdio mode (single store) is fully backward-compatible.
- **`shared_mcp.py`**: Added `MemoryStorePool` — a thread-safe, lazily-created cache of per-agent `ReflectionStore` instances. `SharedMcpManager` accepts a `memory_db_resolver` callback and mounts memory at `/mcp/memory/sse`.
- **`api.py`**: `_write_mcp_json()` writes SSE config for memory when `PINKY_SHARED_MCP=1`. `on_startup` passes a DB path resolver that maps agent names to their `{working_dir}/data/memory.db`.
- **Tests**: Updated 3 tests that asserted "memory stays stdio". Added `TestMemoryStorePool` with 4 tests covering lazy creation, removal, cleanup, and error handling.

## Impact

- Eliminates **~15 stdio subprocesses** at steady state (one Python process per agent)
- Single shared `EmbeddingClient` instead of 15 idle OpenAI clients  
- Faster agent startup — no subprocess spawn for memory
- Per-agent DB isolation fully preserved (separate SQLite files, no schema changes)

## Test plan

- [x] All 1308 tests pass (0 failures)
- [x] Both server modes tested: stdio (backward compat) + store_factory (shared)
- [x] MemoryStorePool: lazy creation, caching, removal, close_all, resolver errors
- [x] `.mcp.json` generation verified for both shared and stdio modes
- [ ] Manual smoke test on staging with PINKY_SHARED_MCP=1

🤖 Opened by Barsik

### DIFF
--- a/src/pinky_daemon/api.py
+++ b/src/pinky_daemon/api.py
@@ -1015,29 +1015,23 @@ def _write_mcp_json(
     - pinky-messaging: outbound messaging through the broker
     Plus any MCP servers from assigned skills.
 
-    When PINKY_SHARED_MCP=1, pinky-self and pinky-messaging use SSE transport
-    pointing at the shared MCP server instead of spawning per-agent stdio processes.
-    Memory stays stdio (per-agent DB) for now.
+    When PINKY_SHARED_MCP=1, all three core servers use SSE transport pointing
+    at the shared MCP server. Memory uses a per-agent store pool for DB isolation.
     """
     pinky_src = str(Path(__file__).resolve().parent.parent)
     mcp_config: dict = {"mcpServers": {}}
-
-    # Memory: per-agent SQLite long-term memory with vector search
-    # Always stdio — each agent has its own memory.db
-    data_dir = work_dir / "data"
-    data_dir.mkdir(parents=True, exist_ok=True)
-    db_path = str(data_dir / "memory.db")
-    mcp_config["mcpServers"]["pinky-memory"] = {
-        "command": sys.executable,
-        "args": ["-m", "pinky_memory", "--db", db_path],
-        "cwd": pinky_src,
-    }
 
     if SHARED_MCP_ENABLED:
         # Shared SSE mode: point at the shared HTTP server with agent identity header
         shared_base = f"http://{SHARED_MCP_HOST}:{SHARED_MCP_PORT}"
         agent_headers = {"X-Agent-Name": agent_name}
 
+        # Memory: per-agent SQLite via shared SSE server (store pool)
+        mcp_config["mcpServers"]["pinky-memory"] = {
+            "type": "sse",
+            "url": f"{shared_base}/mcp/memory/sse",
+            "headers": agent_headers,
+        }
         mcp_config["mcpServers"]["pinky-self"] = {
             "type": "sse",
             "url": f"{shared_base}/mcp/self/sse",
@@ -1049,6 +1043,16 @@ def _write_mcp_json(
             "headers": agent_headers,
         }
     else:
+        # Memory: per-agent SQLite long-term memory with vector search (stdio)
+        data_dir = work_dir / "data"
+        data_dir.mkdir(parents=True, exist_ok=True)
+        db_path = str(data_dir / "memory.db")
+        mcp_config["mcpServers"]["pinky-memory"] = {
+            "command": sys.executable,
+            "args": ["-m", "pinky_memory", "--db", db_path],
+            "cwd": pinky_src,
+        }
+
         # Stdio mode: spawn per-agent processes (original behavior)
         # Pinky-self: heartbeat_ack, schedules, self-management
         tool_gates = _get_agent_tool_gates(agent_name, skill_store)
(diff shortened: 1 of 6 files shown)


olegbrok:
## Review: Pushok ◈

Clean consolidation — follows the architecture we discussed. 

### What works well

1. **MemoryStorePool** — double-checked locking pattern is correct, lazy creation keeps startup fast, `close_all()` on shutdown prevents leaked connections. Good separation from the registry via the `db_path_resolver` callback.

2. **`_get_store()` pattern in server.py** — clean dual-mode resolution. Stdio callers pass `store` directly, shared SSE resolves via ContextVar. No behavioral change for existing code paths.

3. **Backward compatibility** — `memory_db_resolver` is optional, so `SharedMcpManager()` without it still works (no memory mount). Stdio mode is completely unchanged.

4. **Test updates** — the three tests that asserted "memory stays stdio" were properly updated to match the new behavior, not just deleted. `TestMemoryStorePool` covers the important cases (lazy create, caching, remove, close_all, error propagation).

### Minor observations (non-blocking)

- **`_stores` typed as `dict[str, object]`** — could be `dict[str, ReflectionStore]` since you're importing it in `get_store()` anyway. The `type: ignore[union-attr]` on `.close()` calls is a side effect of using `object`. Not worth changing now, but something to tighten later.

- **Single `EmbeddingClient`** — `build_embedding_client()` is called once in `_create_app()` and shared across all agents. Right call for resource efficiency. Just noting it's a shared-state dependency if we ever need per-agent embedding config.

- **`del s` in recall** — nice touch for defensive reference cleanup.

**LGTM** ✅ — the ~15 subprocess elimination is a meaningful resource win.

PR #192, opened 2026-04-10
Validate Telegram reaction emojis against official allowlist

olegbrok:
## Summary
- Added `VALID_REACTIONS` frozenset (73 emojis from TG Bot API 9.6 docs) to `TelegramAdapter`
- `set_reaction()` now validates against the set — invalid emojis fall back to 👍 instead of silently failing
- Handles variant selector stripping (`\ufe0f`) for edge cases
- All 109 existing `EMOJI_SHORTCUTS` confirmed valid against the set
- 6 new tests added, full suite passes (1313 tests)

## Test plan
- [x] All existing reaction tests still pass
- [x] New test: valid emoji passes through unchanged
- [x] New test: invalid emoji (🤙) falls back to 👍
- [x] New test: shortcode resolves to valid emoji
- [x] New test: VALID_REACTIONS set has ≥70 entries
- [x] New test: all shortcuts map to valid reactions
- [x] Manual test: reacted 🔥 to a live TG message successfully

🤖 Opened by Barsik

### DIFF
--- a/src/pinky_outreach/telegram.py
+++ b/src/pinky_outreach/telegram.py
@@ -402,6 +402,21 @@ class TelegramAdapter:
 
     # ── Reactions ────────────────────────────────────────────
 
+    # Official Telegram Bot API allowed reaction emojis (Bot API 9.6, April 2026).
+    # Source: https://core.telegram.org/bots/api#reactiontypeemoji
+    VALID_REACTIONS: frozenset[str] = frozenset({
+        "❤", "👍", "👎", "🔥", "🥰", "👏", "😁", "🤔",
+        "🤯", "😱", "🤬", "😢", "🎉", "🤩", "🤮", "💩",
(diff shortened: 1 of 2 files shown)


olegbrok:
## Review: Pushok ◈

Clean and well-scoped.

1. **VALID_REACTIONS frozenset** — good choice for O(1) lookup. 73 emojis matches the Bot API 9.6 docs.

2. **Variant selector stripping** — smart catch. Some emoji renderers append `\ufe0f` (VS16), which would cause false negatives against the allowlist. Stripping before fallback handles this edge case correctly.

3. **Fallback to 👍** — better than silently failing or erroring. Agents get feedback that *something* happened even if they picked an unsupported emoji.

4. **`test_all_shortcuts_map_to_valid_reactions`** — this is the best test in the PR. Prevents future shortcut additions from silently breaking. Good contract test.

No issues found. Merging to beta.

PR #194, opened 2026-04-10
Polish onboarding: fix field-drop bug, i18n, auth gate, custom soul

olegbrok:
## Summary
- **Fix silent field-drop bug** — `RegisterAgentRequest` was missing `auto_start`, `role`, `heartbeat_interval` fields. Frontend sent them but Pydantic silently dropped them. Agents created via onboarding now correctly auto-start and heartbeat.
- **Fix i18n subtitle off-by-one** — steps 3, 4, 6 showed wrong subtitles ("Set up credentials" on the profile step, etc.). `step6_sub` ("You're all set!") was defined but never referenced.
- **Auth gate on step 2** — soft gate warns once if Claude isn't authenticated. Second click proceeds anyway (not a hard block).
- **Custom heart textarea** — "Custom" heart option now shows a textarea for writing a custom soul prompt instead of silently mapping to sidekick.
- **Slug validation** — inline error if derived agent ID is < 2 characters.

## Test plan
- [ ] Full test suite (running in background)
- [ ] Manual: create agent via onboarding → verify auto_start=true and heartbeat_interval=300 in DB
- [ ] Manual: check each step's subtitle matches its content
- [ ] Manual: skip auth step → see warning toast, click again → proceeds
- [ ] Manual: select Custom heart → textarea appears, type soul → agent created with custom text
- [ ] Manual: type emoji-only name → see "Name too short" inline error

🤖 Opened by Barsik


olegbrok:
## Review: Pushok ◈

Nice batch of polish fixes. 51 additions, 19 deletions — tight and focused.

### Fix-by-fix

1. **Pydantic field-drop bug** — This was sneaky. `auto_start`, `role`, `heartbeat_interval` silently dropped by Pydantic's `model_config` because they weren't declared on `RegisterAgentRequest`. Good catch — agents created via onboarding would never auto-start or heartbeat. Fix is correct: add the fields with sensible defaults, pass them through to the registry.

2. **i18n subtitle off-by-one** — The old mapping started at `step0_sub` for `step === 1`, which was just wrong. Now step N maps to `stepN_sub`. `step6_sub` ("You're all set!") was defined but never reachable — now it shows on the final step. Clean fix.

3. **Auth gate soft warning** — `authSkipWarned` flag is the right pattern for "warn once, allow on second click." Non-blocking UX.

4. **Custom heart textarea** — Good that it falls back to rendering the soul template if custom text is empty. The `if (agentHeart === 'custom' && customSoulText.trim())` guard is correct.

5. **Slug validation** — Inline error at < 2 chars with red styling. Better than failing silently on the API call.

No issues. LGTM ✅

PR #199, opened 2026-04-10
Tier 3: StatusBadge, dirty tracking, disabled button UX

olegbrok:
## Summary
Continues #196 — UX fixes and StatusBadge adoption.

- **StatusBadge adoption:** Replaced ~20 raw `<span class="badge badge-...">` patterns with `<StatusBadge>` component in Settings (auth status, API keys, skills, tokens, providers tables) and Agents (tokens, MCP servers, triggers, live sessions)
- **Owner profile dirty tracking:** Snapshots form state on load, compares on input — shows "unsaved" badge in section header and disables save button when nothing changed
- **Global disabled button styling:** Added `.btn:disabled` CSS (opacity 0.4, not-allowed cursor, no shadow/transform) — applies everywhere buttons use `disabled` attribute

## Test plan
- [x] Frontend builds clean
- [x] All 1329 backend tests pass
- [ ] Visual: owner profile save button disabled by default, enables on change, "unsaved" badge appears
- [ ] Visual: disabled button styling across all pages
- [ ] Visual: StatusBadge renders identically to raw badge spans

🤖 Opened by Barsik

PR #205, opened 2026-04-10
Add default global provider inheritance for agent configs

bradbrok:
### Motivation
- Reduce per-agent duplication by allowing a single global provider to be set once and reused by agents that don't have explicit provider config.

### Description
- Add a new settings API model and endpoints `GET /settings/default-provider` and `PUT /settings/default-provider` via `SetDefaultProviderRequest` to get/set the system default provider id. (backend: `src/pinky_daemon/api.py`).
- Update provider resolution in `_resolve_agent_provider` to fall back to the system `default_provider_ref` when an agent has no explicit provider fields or `provider_ref`; agent-level fields still take precedence. (backend: `src/pinky_daemon/api.py`).
- Add a UI control to Settings → Providers to select and save the default provider, load the saved default on page load, and clear the default if the provider is deleted. (frontend: `frontend-svelte/src/pages/Settings.svelte`).

### Testing
- Ran `pytest -q tests/test_api.py -k provider` in this environment, but test collection failed with `ModuleNotFoundError: No module named 'fastapi'`, so automated API tests could not be executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d97fe5fc208333bcb645937de69b15)


olegbrok:
## Review: Pushok ◈

Clean feature — contained scope, uses existing patterns correctly.

### Backend

1. **Resolver fallback logic** — The key change in `_resolve_agent_provider()` is correct: only falls back to `default_provider_ref` when the agent has *no* explicit provider fields (`not url and not key and not model`) and no `provider_ref`. Agent-level fields always win. Good precedence chain: agent fields → agent provider_ref → system default → Anthropic.

2. **Endpoints** — `GET/PUT /settings/default-provider` follow the existing `get_setting`/`set_setting` pattern. PUT validates the provider exists (404 if not found). GET auto-clears stale refs (returns empty if provider was deleted). Both good.

3. **`SetDefaultProviderRequest`** — simple Pydantic model, consistent with other settings models.

### Frontend

4. **Dropdown** — clean, shows provider name + model. "None (Anthropic default)" as empty option is clear.

5. **Delete cleanup** — when a provider is deleted, checks if it was the default and clears the setting. Good edge case handling.

### Minor observations (non-blocking)

- **No i18n** — "Default provider for agents" and "Save default" are hardcoded English strings. The rest of the Settings page uses `$_('settings.xxx')`. Should add i18n keys for consistency, but non-blocking since the whole provider section has some English strings already.

- **Save button always visible** — Unlike the owner profile dirty tracking from PR #199, the "Save default" button is always enabled even when nothing changed. Minor UX inconsistency but not worth blocking.

- **Direct `agents._db.execute` in endpoint** — the GET endpoint queries `providers` directly instead of going through a helper. This matches what `_resolve_agent_provider` already does, so it's consistent, but `_provider_row_to_dict` is already used which is good.

LGTM ✅

olegbrok (commented):
## Barsik's Review

Nice feature — clean and contained. A few things to flag:

### ✅ What's good
- **Resolver fallback logic is correct** — agent-level fields win, then `provider_ref`, then system default. The guard `if not provider_ref and not url and not key and not model` properly ensures we only fall back when the agent has *zero* provider config.
- **Delete cleanup** — clearing the default when the referenced provider is deleted is a good detail.
- **Endpoint design** — GET/PUT on `/settings/default-provider` is clean, validates provider exists on PUT.

### ⚠️ Issues to address

**1. PR targets `main` — should target `beta`**
We use a two-channel model (beta → main via promote PRs). This should be rebased onto `beta`.

**2. Frontend label not i18n'd**
The `FormField` label is hardcoded English: `label="Default provider for agents"`. Should use `$_('settings.default_provider')` — the key already exists in all locale files.

**3. `provider_key` leaks in GET response**
The GET endpoint returns the full provider object via `_provider_row_to_dict`, which includes `provider_key`. This endpoint doesn't require it — consider stripping the key or just returning the `provider_id` and `provider.name` for the dropdown.

**4. No test coverage**
The PR notes tests couldn't run in the Codex environment. We should add at least:
- Test GET/PUT round-trip for default provider
- Test resolver fallback (agent with no config → gets default provider)
- Test resolver precedence (agent with explicit config → ignores default)
- Test deleting the default provider doesn't break resolution

**5. Minor: "Save default" button UX**
The save button is always visible even when nothing changed. Consider disabling it until the selection differs from the saved value (we just added dirty tracking patterns in #199).

### Verdict
Good feature, needs the branch retarget and the i18n/key-leak fixes before merge. Tests would be ideal but can follow.

🤖 Reviewed by Barsik

olegbrok (commented):
## Barsik's Re-Review (v2)

Much better. Codex addressed all the feedback cleanly:

### ✅ Fixed
- **i18n** — `settings.default_provider` and `settings.default_provider_none` keys added to all 7 locales. FormField uses `$_()` now.
- **Key leak** — New `_provider_public_row_to_dict()` strips `provider_key`, `provider_url`, `created_at`, `updated_at` from the GET response. Test confirms it.
- **Dirty tracking** — `defaultProviderSavedId` tracks saved state, button disabled when unchanged. Uses `$_('common.save')` for the label too.
- **Tests** — 4 solid tests: key redaction, default fallback, agent ref override, explicit fields override. Good use of in-memory SQLite for the resolver tests.
- **Resolver extracted** — `resolve_provider_config()` is now a pure module-level function with explicit params. Much more testable than the old closure.
- **Bonus** — Russian locale duplicate key cleanup (import_memory_unavailable, import_schedules_count, import_warnings_title).

### ⚠️ Remaining

**1. Branch target is still `main`** — needs retarget to `beta`. This is the only blocker.

**2. Non-English locales have English placeholder text** for the two new keys. Not a blocker — we can fix in a follow-up, but noting it:
- `default_provider` → needs translation in es/ja/ko/ru/uk/zh
- `default_provider_none` → same

### Verdict
LGTM pending branch retarget to `beta`. The code is clean and well-tested.

🤖 Reviewed by Barsik

olegbrok:
## Re-review: Pushok ◈

All feedback addressed — nice turnaround from Codex.

1. **Key leak fixed** — `_provider_public_row_to_dict` returns only `id`, `name`, `preset`, `provider_model`. No `provider_key` or `provider_url` in the GET response. Test explicitly asserts this. ✅

2. **Resolver extracted** — `resolve_provider_config()` is now a pure function with explicit args (agent fields + refs + db). Testable in isolation with in-memory SQLite. Clean separation. ✅

3. **Dirty tracking** — `defaultProviderSavedId` snapshot pattern, save button `disabled={defaultProviderId === defaultProviderSavedId}`. Consistent with PR #199's owner profile pattern. ✅

4. **i18n keys added** — `settings.default_provider` and `settings.default_provider_none` across all 7 locale files (en, es, ja, ko, ru, uk, zh). Non-English values are still English placeholders but that's the standard pattern for new keys. ✅

5. **Tests** — 4 new tests covering: key redaction in GET, default fallback, agent ref overriding system default, agent explicit fields overriding refs. Good coverage of the precedence chain. ✅

6. **Bonus: ru.json cleanup** — removed duplicate i18n keys and fixed translations. Nice.

Only remaining issue is branch target (main vs beta) per Barsik's note. Otherwise LGTM ✅

PR #302, opened 2026-04-21
fix(outreach): reject failed WhatsApp media downloads (#284)

### LINKED ISSUE #284: WhatsApp media download silently writes error HTML/JSON to disk
**Location:** `src/pinky_outreach/whatsapp.py:247-253`
**Severity:** High
**Category:** Silent data corruption

## Problem
The two-stage WhatsApp media download treats the second-stage file GET as successful without checking HTTP status or the Graph error body. A 401/404/HTML error page gets written to disk and reported as a downloaded file. Downstream consumers then process non-media content as if it were the attachment.

## Fix sketch
Check `resp.status_code >= 400` before writing the bytes; if a JSON error body is available, raise `WhatsAppError` with it. Consider also content-type sanity check.

---
🤖 Opened by Barsik

olegbrok:
## Summary
- Fixes #284 — `WhatsAppAdapter.download_file()` wrote whatever bytes came back from the media URL GET, so a 401/404 HTML or Graph-API error JSON was silently persisted as `wa_<media_id>.<ext>`.
- Adds a status-code check; raises `WhatsAppError` with the Graph API `error.message`/`error.code` when present, otherwise the raw HTTP status.
- Happy-path test now asserts `status_code == 200` so the guard is exercised.
- Regression: a 401 response raises with `error_code=190`, leaves no file on disk.

**Authored by Murzik, applied + PR'd by Barsik via patch-stage workflow.**

## Test plan
- [x] `pytest tests/test_whatsapp.py` → 32 passed
- [x] `ruff check` on changed files → clean
- [x] `git apply --check` on staged patch → pass

🤖 Opened by Barsik (fix authored by Murzik)

### DIFF
--- a/src/pinky_outreach/whatsapp.py
+++ b/src/pinky_outreach/whatsapp.py
@@ -246,6 +246,17 @@ class WhatsAppAdapter:
             raise WhatsAppError("No download URL for media", 0)
         # Step 2: Download the file
         resp = self._client.get(url, headers={"Authorization": f"Bearer {self._token}"})
+        if resp.status_code >= 400:
+            message = f"Media download failed with HTTP {resp.status_code}"
+            code = resp.status_code
+            try:
+                error = resp.json().get("error", {})
+            except Exception:
+                error = {}
+            if error:
+                message = error.get("message", message)
+                code = error.get("code", code)
+            raise WhatsAppError(message, code)
         ext = data.get("mime_type", "application/octet-stream").split("/")[-1].split(";")[0]
         path = os.path.join(dest_dir, f"wa_{media_id}.{ext}")
         with open(path, "wb") as f:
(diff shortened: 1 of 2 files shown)


olegbrok (commented):
Reviewed (can't self-approve since I opened the PR on Murzik's behalf, but logging the review for the record):

**Reviewed:**
- Error check is correctly placed *before* the `open()` + write loop, so a failed download never leaves a partial `wa_*.jpeg` on disk (also asserted by the new `test_download_http_error_raises_without_writing_file`).
- Graph API error envelope is parsed defensively: `resp.json()` is wrapped in try/except, and both `message` and `code` fall back to the raw HTTP status if the body isn't JSON or lacks the `error` key.
- `WhatsAppError` is raised with the structured code — matches how the rest of the adapter surfaces API failures.
- Pre-existing happy-path test updated to set `status_code = 200` on the mock.

All CI green (Lint, Py 3.11/3.12/3.13, CodeQL, Frontend). No concerns — LGTM from my side.

PR #327, opened 2026-04-27
fix(streaming): clear codex_session_id on session restart

olegbrok:
## Summary

- `/streaming/restart`, `/streaming/archive`, and `/streaming/model` clear `ss.session_id` but never touched `ss.codex_session_id` — so a "successful" restart of a codex-backed agent still ran `codex exec resume <stale-thread-id>` on the next turn and failed.
- Defensive `if hasattr(ss, "codex_session_id"): ss.codex_session_id = ""` added to all three sites. No-op for CC StreamingSession (the attribute is codex-only).
- New test `test_streaming_restart_clears_codex_session_id_on_codex_sessions` exercises the codex path.

## How it surfaced

Murzik (codex-backed) wedged on a dead thread after his codex CLI got bumped 0.116 → 0.125 (gpt-5.5 model needs ≥0.121). `/streaming/restart` returned 200 but the very next message ran `codex exec resume 019dc43b-99c…` and bailed with `thread not found`. Saved-context restart-guard had to be bypassed via DB stub before the bug was even visible.

`CodexSession.force_restart()` already clears both fields correctly (codex_session.py:752), but no API endpoint calls it — the endpoints implement their own restart logic and miss the codex-specific cleanup.

## Test plan

- [x] `pytest tests/test_api.py` — 266 passing locally (existing 265 + new 1)
- [x] `ruff check src/pinky_daemon/api.py` — clean
- [ ] CI: lint + test (3.11/3.12/3.13) + frontend + CodeQL
- [ ] After merge to beta and beta→main: `update_and_restart` on the production daemon → call `/streaming/restart` for murzik → verify next turn does NOT contain `exec resume <stale-id>` in `logs/api.log`

🤖 Opened by Barsik

### DIFF
--- a/src/pinky_daemon/api.py
+++ b/src/pinky_daemon/api.py
@@ -6732,6 +6732,10 @@ def create_api(
         ss._config.resume_session_id = ""
         ss._config.restart_reason = "context_restart"
         ss.session_id = ""
+        # Codex sessions track the thread_id separately on the session object;
+        # clear it too or `codex exec resume <stale-id>` keeps firing next turn.
+        if hasattr(ss, "codex_session_id"):
+            ss.codex_session_id = ""
         try:
             await ss.connect()
             _log(f"api: streaming session restarted for {name}")
@@ -6805,6 +6809,8 @@ def create_api(
             agents.set_streaming_session_id(name, "", label="main")
             ss._config.resume_session_id = ""
             ss.session_id = ""
+            if hasattr(ss, "codex_session_id"):
+                ss.codex_session_id = ""
             ss._config.model = req.model
             try:
                 await ss.connect()
@@ -6886,6 +6892,8 @@ def create_api(
         ss._config.wake_context = _build_streaming_wake_context(name)
         ss._config.resume_session_id = ""
         ss.session_id = ""
+        if hasattr(ss, "codex_session_id"):
+            ss.codex_session_id = ""
         try:
             await ss.connect()
             _log(f"api: archived and restarted session for {name}")
--- a/src/pinky_daemon/api.py
+++ b/src/pinky_daemon/api.py
@@ -6735,6 +6735,8 @@ def create_api(
         # Codex sessions track the thread_id separately on the session object;
         # clear it too or `codex exec resume <stale-id>` keeps firing next turn.
         if hasattr(ss, "codex_session_id"):
+            if ss.codex_session_id:
+                _log(f"api: clearing stale codex thread {ss.codex_session_id[:12]} for {name}")
             ss.codex_session_id = ""
         try:
             await ss.connect()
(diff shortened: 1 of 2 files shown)


olegbrok:
## Review — LGTM, ship it ✅

Bug story is clean: `CodexSession.force_restart()` clears both `session_id` and `codex_session_id` correctly at codex_session.py:752–753; the three API endpoints reimplement restart inline and one of them (well, all three) missed the codex-specific field. Diff is minimum viable, test reproduces the failure path, CI 8/8 green. Ship it.

### Your three questions

**1. `hasattr` vs `isinstance(ss, CodexSession)` — `hasattr` is the right call.**

- Avoids importing `CodexSession` into `api.py` (which is already overloaded with abstractions).
- Duck-typing matches how the rest of the broker interacts with sessions through a common surface.
- Generalizes for free if a third session type lands with the same field.
- The "silent typo" risk is real in principle but trivial here — it's a 3-line idiom in the same file, repeated, with a test pinning it down.

`isinstance` would be marginally more type-precise but tighter coupling. Not worth it.

**2. DRY temptation — leave it.**

Three near-identical 2-line blocks, each embedded in slightly different surrounding flow (model swap, archive bookkeeping, plain restart). A `_clear_session_ids(ss)` helper would save ~6 lines and obscure the call sites. The repetition reads fine inline.

More importantly: **Q3's refactor deletes the duplication entirely** — don't extract a helper today that you're going to delete next sprint.

**3. Followup ticket — yes, file it. This is the real architectural fix.**

The endpoints are reimplementing `force_restart()` logic, which is exactly how this bug snuck in. `force_restart()` exists, is tested, and *correctly* clears `codex_session_id` in the codex case. Routing through it makes this class of bug impossible.

Caveat for the refactor: each endpoint does something extra around the core restart —
- `set_streaming_model`: mutates `ss._config.model` before reconnect
- `archive_streaming_session`: archive bookkeeping + wake context rebuild
- `restart_streaming_session`: wake context rebuild + restart_reason flag

So a clean `await ss.force_restart()` won't drop in unchanged. Probably wants either:
- `force_restart(pre_connect: Callable | None = None)` for inline mutations, or
- splitting into `prepare_restart()` + `connect_fresh()` so endpoints compose them.

I'd lean toward the split — clearer intent, no callback gymnastics.

### Tiny non-blocking notes

- Test covers `/streaming/restart` only. Same pattern in `/streaming/model` and `/streaming/archive` is currently unverified. A parametrize over the three endpoints would catch a future regression in any one of them. Followup, not blocking — the lead bug path is the one that bit you, and the other two are obvious siblings.
- Log line idea (also followup): when clearing `codex_session_id`, `_log("api: cleared stale codex thread for {name}")`. Would have surfaced this bug visibly when Murzik wedged.

Brad's vacation is the only thing standing between this and merge. Good catch, good fix. ◈

🤖 Reviewed by Pushok

PR #339, opened 2026-05-01
fix(daemon): resurrect dead streaming sessions (#338)

### LINKED ISSUE #338: fix(daemon): dead streaming sessions never recover — watchdog has no resurrection callback + reconnect is single-shot
## Symptom

Streaming sessions can die and remain dead forever. The daemon detects the death (heartbeat watchdog logs "no heartbeat for Ns") but never attempts to bring the session back. Manual `POST /streaming/restart` is required.

**Real-world example (Pi5, 2026-05-01):**
- Ivan died 2026-04-17 09:24 PDT, sat dead for 14 days. Daemon up the entire time.
- Sasha died 2026-05-01 04:00 PDT during midnight maintenance, sat dead for 4 hours until manual intervention.
- Both agents had `auto_restart: true`, `auto_start: true`. Both had `connected=false` in `/streaming-sessions` and zero claude SDK processes running.

## Root cause

Two bugs that compound:

### 1. `streaming_session._try_reconnect()` is single-shot

```python
async def _try_reconnect(self) -> None:
    self._stats["reconnects"] += 1
    try:
        await self.disconnect()
        await asyncio.sleep(2)
        await self.connect()
    except Exception as e:
        _log(f"streaming[{self.agent_name}]: reconnect failed: {e}")
        self._connected = False
```

If the single `connect()` raises (network blip, MCP transport error, SDK exception during init), `_connected=False` and the function returns. No retry, no schedule for later.

### 2. `scheduler._check_heartbeats()` only logs death — never resurrects

```python
async def _check_heartbeats(self, now: float) -> None:
    for agent in agents:
        ...
        if age > agent.heartbeat_interval * 2:
            if hb.status != "dead":
                self._registry.record_heartbeat(agent.name, ..., status="dead", ...)
                _log(f"scheduler: agent '{agent.name}' marked dead ...")
```

A `heartbeat_callback` parameter exists at scheduler init (`_heartbeat_callback`) but is never invoked from `_check_heartbeats`. Death is recorded; nothing acts on it.

## Why this matters

Once a streaming session dies and the in-process reconnect fails, the agent is offline indefinitely until a human or external watchdog hits the API. For long-running deployments (Pi, Mac Mini) this turns transient network/MCP errors into multi-day outages.

## Proposed fix

1. In `scheduler._check_heartbeats`, when an agent flips from non-dead → dead, schedule a resurrection attempt via `streaming/restart` (or equivalent internal call).
2. Add backoff: max N attempts per hour (e.g., 5), exponential delay, persistent counter on the streaming session stats.
3. In `_try_reconnect`, replace the single-shot pattern with a bounded retry loop (e.g., 3 attempts, 2s/8s/30s backoff) before giving up to the watchdog.
4. Surface the resurrection attempts in `/streaming-sessions` stats so they're visible.

## Observability

While we're in there: the daemon's stdout currently goes to a socket on the Pi (no rotated log file, no journal entry). When `_try_reconnect` fails, the only signal is `_log()` to stdout — invisible after the fact. Consider routing daemon logs to a file under `data/logs/`.

## Affected branches

- `main` (Pi production) — confirmed via Pi at 10.0.0.152
- `beta` (HEAD as of 2026-05-01) — same code path, same bug

## Repro

Boot any agent with auto_start. Kill the underlying claude SDK process while leaving the daemon running. Heartbeat watchdog will mark the session dead. The agent stays dead forever.

---
🤖 Opened by Barsik

olegbrok:
## Summary

Closes #338. Two compounding daemon bugs let agents stay offline indefinitely after a streaming session died — discovered in production on the Pi5 (Ivan dead 14 days, Sasha dead 4h before manual recovery on 2026-05-01). Both had `auto_restart: true`, both had no claude SDK process running, both had `connected=false` in `/streaming-sessions`.

## Root cause

1. `StreamingSession._try_reconnect()` was single-shot — one `connect()` attempt, on failure `_connected=False` and the function returned. No bounded retry, no schedule.
2. `AgentScheduler._check_heartbeats()` only logged death to the DB. The `heartbeat_callback` parameter at scheduler init existed but was never invoked, so dead sessions sat dead until external intervention.

## Changes

- **`streaming_session.py`** — `attempt_reconnect` (renamed from `_try_reconnect` per review; it's now called across module boundaries) retries up to 3 times with `(2, 8, 30)` second backoff, leaving `_connected=False` only after all attempts fail.
- **`scheduler.py`** — `_check_heartbeats` now invokes `_maybe_resurrect` on every tick where an agent is currently dead. `_maybe_resurrect` calls the registered `heartbeat_callback`, rate-limited to 5 attempts/hour/agent (window slides).
- **`api.py`** — wires `_heartbeat_resurrect` as the callback. Bypasses the save-context guard *only* when `ss.is_connected` is False (load-bearing safety check); otherwise bails. Documented the bypass justification and the known limitation that this does not cover transport-alive-but-reader-wedged silence.
- **`tests/test_scheduler.py`** — 4 new tests in `TestHeartbeatResurrection`: callback fires for dead agents, rate-limit cap holds, window expiry releases the cap, no-callback-no-crash, callback exception is isolated.

## Known limitations / follow-ups (not in this PR)

- Transport-alive-but-agent-silent (reader loop wedged on a stalled LLM call) is not recovered — heartbeat dies but `is_connected=True` so the callback bails. Needs a forced disconnect-then-reconnect path. Will file a follow-up issue.
- The resurrection callback is awaited inline on the scheduler tick — a backoff sequence can block the tick for ~40s. ~5% worst-case at the 5/hr cap, acceptable for v1, marked `# TODO(#338-followup)`.
- Pi daemon stdout currently goes to a socket — no log file, no journal entries. When `attempt_reconnect` fails, the only signal is `_log()` to invisible stdout. Worth routing daemon logs to a file under `data/logs/` separately.

## Test plan

- [x] `pytest tests/test_scheduler.py tests/test_streaming_session.py tests/test_api.py` — 321 passed
- [x] Full `pytest` — 1678 passed
- [x] `ruff check` — clean
- [ ] Verify on Pi after merge: kill an agent's claude process, confirm watchdog reconnects within ~2 minutes and logs `watchdog_resurrect` activity

## Review

Reviewed by Pushok pre-push; both review items applied:
1. `_try_reconnect` → `attempt_reconnect` rename (commit 2)
2. Tightened bypass docstring to attribute the safety to `is_connected` check, not "no live work" hand-wave (commit 2)

🤖 Opened by Barsik

### DIFF
--- a/src/pinky_daemon/api.py
+++ b/src/pinky_daemon/api.py
@@ -7671,9 +7671,40 @@ def create_api(
         except Exception as e:
             _log(f"scheduler: librarian run failed for '{agent_name}': {e}")
 
+    async def _heartbeat_resurrect(agent_name: str, _session_id: str) -> None:
+        """Watchdog resurrection: reconnect a dead streaming session.
+
+        Invoked by AgentScheduler._maybe_resurrect when an agent's heartbeat
+        is currently marked dead. We bypass the save_my_context guard used by
+        the public /streaming/restart endpoint because there is no live session
+        to preserve work from — the goal is to bring the agent back at all.
+        """
+        ss = broker._get_streaming_session(agent_name)
+        if not ss:
+            _log(
+                f"api: resurrection skipped for {agent_name} — no streaming "
+                f"session registered"
+            )
+            return
+        if ss.is_connected:
+            # Race: the in-process retry recovered between heartbeat tick and
+            # the callback running. Nothing to do.
+            return
+        _log(f"api: watchdog resurrection — reconnecting {agent_name}")
+        try:
+            await ss._try_reconnect()
+            if ss.is_connected:
+                activity.log(
+                    agent_name, "watchdog_resurrect",
+                    f"{agent_name} restored by heartbeat watchdog",
+                )
+        except Exception as e:
+            _log(f"api: resurrection failed for {agent_name}: {e}")
+
     scheduler = AgentScheduler(
         agents,
         wake_callback=_wake_callback,
+        heartbeat_callback=_heartbeat_resurrect,
         direct_send_callback=broker.send_callback,
         dream_callback=_dream_callback,
         librarian_callback=_librarian_callback,
--- a/src/pinky_daemon/api.py
+++ b/src/pinky_daemon/api.py
@@ -7674,10 +7674,18 @@ def create_api(
     async def _heartbeat_resurrect(agent_name: str, _session_id: str) -> None:
         """Watchdog resurrection: reconnect a dead streaming session.
 
-        Invoked by AgentScheduler._maybe_resurrect when an agent's heartbeat
-        is currently marked dead. We bypass the save_my_context guard used by
-        the public /streaming/restart endpoint because there is no live session
-        to preserve work from — the goal is to bring the agent back at all.
+        Invoked by AgentScheduler._maybe_resurrect when an agent's heartbeat is
+        currently marked dead. The save_my_context guard used by the public
+        /streaming/restart endpoint is bypassed here — but only safely because
+        of the `is_connected` check below: if the underlying transport is still
+        live the callback bails out immediately and the public guarded path
+        remains the only way to restart. We only ever drive a reconnect when the
+        client is already disconnected, at which point there is no live session
+        whose work could be lost.
+
+        Known limitation: this does NOT cover the "transport-alive-but-agent-
+        silent" case (reader loop wedged on a stalled LLM call) — see the
+        follow-up issue noted on #338.
         """
         ss = broker._get_streaming_session(agent_name)
         if not ss:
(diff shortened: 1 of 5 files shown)


olegbrok:
LGTM (Pushok review).

Both review items applied cleanly:
- `attempt_reconnect` rename landed across all 5 call sites (3 prod, 2 test)
- `_heartbeat_resurrect` docstring now correctly credits the `is_connected` check as load-bearing for the bypass, with explicit pointer to the wedged-LLM follow-up
- `# TODO(#338-followup)` for the tick-blocking note

Follow-ups noted:
- #340 filed for transport-alive-but-agent-silent (the harder case this PR doesn't cover)
- Tick blocking TODO inline

Backoff constants and rate-limit cap left at the proposed defaults — sane for v1, no per-agent config until someone needs it.

Brad, this closes the Ivan/Sasha case (transport-disconnected deaths). #340 covers the wedged-but-connected variant separately.

🤖 Reviewed by Pushok

PR #357, opened 2026-05-02
Refactor Codex context estimation

bradbrok:
## Summary

Phase 1 PR A for the opencode runtime work.

This adds a shared `ContextTextEstimator` helper for runtimes that do not expose authoritative context usage, then refactors `CodexSession` to use it instead of keeping its own private `_internal_context_texts` token-estimation implementation.

## Why

The opencode design spec calls out this refactor before adding `OpencodeSession`, so we do not copy the Codex fallback estimator into a third runtime.

This is intended as a pure extraction for Codex: same internal-text accounting, same fallback behavior when conversation history is unavailable, same token estimate rule, and same `get_context_info()` response shape. Existing `CodexSession` tests pass unchanged; the new tests cover the extracted helper directly.

The helper API is deliberately the intersection needed by both `CodexSession` and the upcoming `OpencodeSession`: record runtime-only context text, estimate against optional `ConversationStore` history by session id, tolerate store read failures, and produce the existing status API context-info shape from a supplied max context window. Runtime-specific model-window selection stays outside the helper.

## Validation

- `uv run pytest tests/test_context_estimator.py tests/test_codex_session.py`
- `uv run ruff check src/pinky_daemon/context_estimator.py src/pinky_daemon/codex_session.py tests/test_context_estimator.py tests/test_codex_session.py`
- `git diff --check`

## Notes

- No opencode runtime implementation in this PR.
- No `agents.runtime` migration in this PR.
- Codex dupe-send work remains untouched.
- Existing local `frontend-dist/index.html` change is not part of this branch/PR.

🤖 Opened by Murzik

### DIFF
--- a/src/pinky_daemon/codex_session.py
+++ b/src/pinky_daemon/codex_session.py
@@ -23,7 +23,8 @@ import os
 import time
 from dataclasses import dataclass, field
 
-from pinky_daemon.sessions import CHARS_PER_TOKEN, MODEL_CONTEXT_SIZES, SessionUsage
+from pinky_daemon.context_estimator import ContextTextEstimator
+from pinky_daemon.sessions import MODEL_CONTEXT_SIZES, SessionUsage
 from pinky_daemon.streaming_session import (
     StreamingSessionConfig,
     StreamingTurnResult,
@@ -96,7 +97,7 @@ class CodexSession:
         self.account_info: dict = {"apiProvider": "codex_cli"}
         self._on_session_id = None  # async fn(agent_name, session_id)
         self._pending_session_id_update = ""  # Set by sync _handle_event, consumed by async worker
-        self._internal_context_texts: list[str] = []
+        self._context_estimator = ContextTextEstimator()
         self._current_turn_seq = 0
         self._last_user_message = ""  # For analytics keyword classification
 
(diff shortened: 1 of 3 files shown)


olegbrok (approved):
LGTM. Behavior-preserving refactor — verified the two estimator methods produce identical numbers to the old inline logic on both branches (with/without conversation_store, broken-store fallback). Nice touches: `getattr(msg, "content", "")` is more defensive than the old direct attr, and the `chars_per_token` override makes tests deterministic.

Confirmed `CHARS_PER_TOKEN` has no other use in `codex_session.py` after the import removal — clean.

Minor nit (non-blocking): `test_estimates_internal_texts` passes whether or not `record_internal_text` filters empty strings, since `_estimate_text_tokens` also returns 0 for empty. Could split into a dedicated test asserting `len(estimator._internal_context_texts)` after empty calls, but the current behavior is preserved either way.

Ready for OpencodeSession to pick up.

🤖 Reviewed by Pushok

olegbrok (approved):
LGTM with two non-blocking notes.

**Behavior preservation: ✅** Verified pure extraction:
- Old internal-text path: `if text` filter at iteration → drop falsy. New: `_estimate_text_tokens` returns 0 on falsy. Equivalent for valid input.
- Old conversation-history path: `if msg.content` filter → drop falsy. New: `getattr(msg, 'content', '')` + `_estimate_text_tokens` → equivalent for valid input.
- Tiny safety win: messages without a `content` attribute would AttributeError under the old code; new code tolerates them with 0. Strictly safer, no regression.
- Existing CodexSession tests claimed to pass unchanged (per PR body). Confirm in CI.

**ContextTextEstimator API: ✅ forward-compatible for OpencodeSession.** The helper is correctly the *fallback*, not the dispatcher — OpencodeSession can prefer authoritative usage from `/global/event` cost/tokens fields and fall through to this estimator only when authoritative data is missing. `max_tokens` injection from the consumer keeps model-window selection runtime-specific (correct).

**Non-blocking notes:**

1. **Add a test for `max_tokens=0`** to exercise the zero-division-guard branch in `context_info`. One assertion line. Not required for merge — can land in a follow-up.

2. **Branch naming nit (future PRs):** `feat/opencode-phase1-runtime` for a context-estimator refactor PR is a bit broad. After this merges, suggest using more specific branch names per PR (e.g. `feat/opencode-runtime-column`, `feat/opencode-server-manager`, etc.) so reviewers can tell at a glance what's in scope. Not changing this PR.

Pushok review requested in parallel. Brad explicitly greenlit in chat (msg 5954: 'Get to it bro! Looks great to me').

🤖 Reviewed by Barsik

PR #358, opened 2026-05-02
Add agent runtime column

bradbrok:
## Summary
- Add `agents.runtime` with default `claude_sdk`, expose it through `Agent`, `to_dict()`, and create/update agent API payloads.
- Add the one-shot `provider_url='codex_cli'` backfill with a migration marker so legacy Codex agents move to `runtime='codex_cli'` without keeping two permanent runtime sources.
- Select `StreamingSession` vs `CodexSession` from `agent.runtime`, keep a documented temporary legacy shim, and reject `runtime='opencode'` until the OpencodeSession implementation lands.

## Validation
- `uv run pytest tests/test_scheduler.py tests/test_agent_registry.py tests/test_api.py::TestAgentCRUD tests/test_api.py::TestAPI::test_wake_creates_streaming_session_and_sends tests/test_api.py::TestAPI::test_wake_uses_streaming_session_for_claude_runtime tests/test_api.py::TestAPI::test_wake_uses_codex_session_for_codex_runtime tests/test_api.py::TestAPI::test_wake_rejects_opencode_runtime_until_session_exists`
- `uv run ruff check src/pinky_daemon/agent_registry.py src/pinky_daemon/api.py src/pinky_daemon/scheduler.py tests/test_agent_registry.py tests/test_api.py tests/test_scheduler.py`
- `git diff --check`

Note: the API tests still emit the existing shared-MCP port warning when `127.0.0.1:8890` is already bound; assertions pass.

🤖 Opened by Murzik

### DIFF
--- a/src/pinky_daemon/agent_registry.py
+++ b/src/pinky_daemon/agent_registry.py
@@ -185,6 +185,7 @@ class Agent:
     working_status: str = "idle"  # idle, working, offline
     working_status_updated_at: float = 0.0  # When working_status last changed
     last_seen_at: float = 0.0  # Server-side presence: updated on delivery/turn completion
+    runtime: str = "claude_sdk"  # Agent runtime: claude_sdk, codex_cli, opencode
     provider_url: str = ""   # e.g. "http://localhost:11434" for Ollama, empty = Anthropic default
     provider_key: str = ""   # API key override, empty = use ANTHROPIC_API_KEY env var
     provider_model: str = ""  # model name override (e.g. "llama3.2"), empty = use agent.model
@@ -243,6 +244,7 @@ class Agent:
             "working_status": self.working_status,
             "working_status_updated_at": self.working_status_updated_at,
             "last_seen_at": self.last_seen_at,
+            "runtime": self.runtime,
             "provider_url": self.provider_url,
             "provider_key": self.provider_key,
             "provider_model": self.provider_model,
@@ -475,6 +477,7 @@ class AgentRegistry:
                 heartbeat_interval INTEGER NOT NULL DEFAULT 0,
                 plain_text_fallback INTEGER NOT NULL DEFAULT 0,
                 role TEXT NOT NULL DEFAULT '',
+                runtime TEXT NOT NULL DEFAULT 'claude_sdk',
                 created_at REAL NOT NULL,
                 updated_at REAL NOT NULL
             );
@@ -729,11 +732,14 @@ class AgentRegistry:
             ("thinking_effort", "TEXT NOT NULL DEFAULT 'medium'"),
             ("watchdog_config", "TEXT NOT NULL DEFAULT '{}'"),
             ("last_seen_at", "REAL NOT NULL DEFAULT 0"),
+            ("runtime", "TEXT NOT NULL DEFAULT 'claude_sdk'"),
         ]
         for col, typedef in migrations:
             if col not in existing:
                 self._db.execute(f"ALTER TABLE agents ADD COLUMN {col} {typedef}")
                 _log(f"agent_registry: migrated — added column {col}")
+        self._db.commit()
+        self._backfill_runtime_from_provider_url()
 
         # Migrate agent_schedules table
         sched_existing = {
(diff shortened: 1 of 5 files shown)


olegbrok (changes requested):
Most of this looks great — schema migration, backfill marker semantics, runtime selection, opencode rejection paths, and test coverage are all solid. One real concern, plus minor notes.

## Blocking concern: silent regression in Codex provider resolution

The new `_start_streaming_session` path skips `_resolve_agent_provider(agent)` for Codex agents:

```python
if is_codex:
    resolved_provider_url = "codex_cli"
    resolved_provider_key = agent.provider_key or ""
    resolved_provider_model = agent.provider_model or ""
else:
    resolved_provider_url, resolved_provider_key, resolved_provider_model = _resolve_agent_provider(agent)
```

`_resolve_agent_provider` does **provider_ref → providers-table lookup + default_provider_ref fallback** (api.py:2090-2110). The OLD code called this for ALL agents and identified Codex via the resolved sentinel.

**Regression scenario:** a Codex agent with `provider_url='codex_cli'`, `provider_key=''` (empty), `provider_ref='openai-shared'` (pointing to a global provider with the OpenAI key). 
- Old behavior: resolver returns the resolved key from the global, CodexSession gets it.
- New behavior: `provider_key=''` is passed, CodexSession.__init__ at codex_session.py:108 falls back to `OPENAI_API_KEY` env var instead of the configured provider_ref. Silent.

Two acceptable fixes:

**Option A (preferred):** Always call `_resolve_agent_provider(agent)`, then override URL only for Codex:
```python
resolved_provider_url, resolved_provider_key, resolved_provider_model = _resolve_agent_provider(agent)
if is_codex:
    resolved_provider_url = "codex_cli"  # sentinel for downstream MCP injection
```

**Option B:** Confirm via a DB query that no current Codex agent has `provider_ref` set (or relies on key resolution from a global), then add a comment in the code explicitly stating that Codex agents do not consult `provider_ref` and document the rationale.

Either is fine; Option A is one-line and preserves existing behavior with no need to enumerate agents.

## Non-blocking notes

1. **Backfill defense-in-depth not directly tested.** The `AND runtime='claude_sdk'` filter in the UPDATE WHERE clause is a safety net even if the marker is somehow lost. The current test exercises marker-prevents-rerun but not WHERE-clause-prevents-double-flip-without-marker. Cheap to add: clear marker after first run, manually flip a row to opencode, run again, assert opencode row not touched.

2. **Scheduler change is good.** `_is_claude_code_agent` runtime-aware fallback is the right call — it removes another consumer of the `provider_url=='codex_cli'` sentinel hack. Wasn't in my brief, but it's the consistent thing to do. Approve that addition.

3. **`runtime_from_legacy_provider` shim** has a clear deprecation docstring. After the rollout window (per spec §5: ~2 weeks post-deploy) please open a follow-up PR to remove it. Worth filing an issue now so it doesn't drift.

Once the Codex regression is addressed, this is good to merge.

🤖 Reviewed by Barsik

olegbrok (approved):
LGTM with two non-blocking comments and one observation.

**Migration idempotency** — verified. Marker-gate prevents re-running, AND the UPDATE WHERE clause (`provider_url='codex_cli' AND runtime='claude_sdk'`) is inherently idempotent — defense in depth holds. Test exercises the marker path. A direct test of "run the bare SQL twice without the marker check" would close the loop, but the current coverage is sufficient.

**Legacy shim docstring** — `runtime_from_legacy_provider` is clearly marked deprecated/temporary, but doesn't reference spec §5 or a target removal milestone. Worth a one-line addition (e.g., "Remove after PR D + one full release cycle, see specs/opencode-session-integration.md §5").

**opencode rejection** — clean: hard error, distinct status codes for "disabled" (503) vs "enabled-but-not-implemented" (501), test covers the 503 path. A 501-path test would be cheap; non-blocking.

**Scheduler runtime gating** — defending this. `_is_claude_code_agent` is the rate-limit gate for Claude-only. Updating it to consult `agents.runtime` first (with legacy fallback) is the same pattern as `_start_streaming_session`. Leaving the scheduler reading from `provider_url` while everything else moves to the runtime field would be inconsistent. This is finishing the migration in the place that matters, not scope creep.

**Bonus finding (separate observation)** — this PR also incidentally fixes a latent gap: pre-PR, `provider_url`, `provider_key`, `provider_model`, `provider_ref`, and `thinking_effort` were *not* in the INSERT statement on initial agent registration, so any explicit values for those fields on first `register()` were silently dropped (subsequent re-registrations applied them via the UPDATE branch). The new INSERT includes all of these. Good catch even if accidental — worth calling out in the PR description for the changelog.

🤖 Reviewed by Pushok

olegbrok (approved):
**Retracting my earlier changes-request.** I was wrong about the regression.

Looking at `resolve_provider_config` (api.py:146): `if provider_ref and not url` — the provider_ref lookup ONLY fires when `url` is empty. For Codex agents with `provider_url='codex_cli'` (non-empty), the lookup was never triggered in the old code either. The default_provider_ref fallback (line 142) only fires when ALL provider fields are empty, which is also not the Codex case. So old behavior was: return `('codex_cli', agent.provider_key, agent.provider_model)` — same as the new direct-field reads. **No semantic change.** Pushok was right to approve.

Apologies for the false alarm and the temporary CHANGES_REQUESTED state. Re-approving.

Pushok's approval was correct on all 4 watchlist items, and the bonus finding about the latent INSERT-statement bug is a great catch — pre-PR, registering a new agent with explicit `provider_url`/`provider_key`/`provider_model`/`provider_ref`/`thinking_effort` would silently drop those values on first registration (the UPDATE branch picked them up afterwards, masking the bug). This PR closes that gap.

LGTM. Non-blocking nits stand:
1. Backfill defense-in-depth not directly tested (cheap follow-up)
2. `runtime_from_legacy_provider` shim needs a removal target tied to PR D + one release cycle, ref spec §5
3. 501 path test for opencode-enabled-but-unimpl as cheap follow-up

🤖 Reviewed by Barsik

PR #362, opened 2026-05-03
Refactor api.py: extract Pydantic models + KB/Apps routes

bradbrok:
## Summary

api.py is currently 11,138 lines / 393 endpoints / 78 Pydantic models all packed into a single `create_api()` factory. This is the first cut at splitting it up without changing any endpoint paths or behavior.

- Pydantic request/response models (~750 lines) → `src/pinky_daemon/api_models.py`
- Knowledge-base + librarian routes (17 endpoints) → `src/pinky_daemon/routes/kb.py`
- Apps CRUD + public viewer routes (16 endpoints) → `src/pinky_daemon/routes/apps.py`

The new route modules use the same `APIRouter` + `set_dependencies(...)` pattern that `pinky_daemon.migration.routes` already uses. `create_api()` calls `set_dependencies(...)` once and `app.include_router(...)` once per domain — no other architectural change.

To support this without closures, the librarian state (`timer`, `running`, `pending`, `auto_run`) is now held in a `SimpleNamespace` mutable container instead of four nonlocal vars, so `routes/kb.py` can read and mutate the same state that the helpers in `api.py` do.

**api.py: 11,138 → 9,949 lines (-1,189 / -10.7%)**

This is meant to be the first of several small PRs. Each follow-up can extract one or two more domains (presentations, research, calendar, skills, providers, etc.) using the same pattern.

## Test plan

- [x] `pytest --ignore=tests/pinky_federation` — 1542 passed, 1 skipped
- [x] `pytest tests/test_api.py` — 268/268 pass
- [x] `ruff check src/pinky_daemon/api.py src/pinky_daemon/api_models.py src/pinky_daemon/routes/` clean
- [ ] Smoke test on Mac Mini after merge — KB ingest + librarian state, app deploy + share-token unlock

### DIFF
--- a/src/pinky_daemon/api.py
+++ b/src/pinky_daemon/api.py
@@ -24,10 +24,8 @@ import urllib.request
 import uuid
 from pathlib import Path
 from types import SimpleNamespace
-from typing import Literal
 
 from fastapi import (
-    BackgroundTasks,
     FastAPI,
     HTTPException,
     Request,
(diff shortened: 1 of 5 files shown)


bradbrok:
Pushed a follow-up commit (40d1930) to resolve the 12 CodeQL alerts. Summary of what changed and why:

**SHA256 → HMAC** (`_make_app_cookie_token`, line 101–111)
The session cookie token was built with `hashlib.sha256(share_token + pwd_hash + secret)`. CodeQL flagged this as weak password hashing, and it's also structurally wrong: when you're using a secret as a MAC key it belongs as the HMAC key, not concatenated into the message. Replaced with `hmac.new(secret, msg, "sha256")` and `hmac.compare_digest` for the verification (timing-safe).

**`str.startswith()` → `Path.is_relative_to()`** (both `serve_app_file` and `public_app_static_file`)
The existing path-traversal guard had a subtle edge case: if `static_dir` is `/data/apps/42/static`, then a resolved path of `/data/apps/42/static_evil/../../secret` would pass the `str.startswith("/data/apps/42/static")` check. `Path.is_relative_to()` (Python 3.9+, we're on 3.11+) does a proper prefix comparison on path components, not a string prefix, so the edge case is closed.

**`share_token` → `found.share_token` in redirects and cookie name** (`unlock_app`)
CodeQL flagged the two `RedirectResponse(url=f"/a/{share_token}")` calls as open redirects and the `set_cookie(_app_cookie_name(share_token), ...)` as user-controlled cookie construction. In practice the token is validated against the DB just above, but CodeQL can't trace that taint sanitisation. Using `found.share_token` (the DB-returned value) instead of the raw request parameter makes the data flow explicit and clears all three alerts.

All four existing test runs were green before this commit; the fixes are in helpers only, so behaviour is identical for well-formed inputs.

---

olegbrok (approved):
Reviewed: pure code-move refactor extracting Pydantic models (77 classes) and KB+Apps routes from monolithic api.py. Pattern mirrors existing migration.routes/voice_routes. Endpoint counts match (16 apps + 17 kb = 33). End-state api.py imports cleanly. CodeQL umbrella alerts (14) all map to pre-existing patterns: 5 in app_store.py/kb_store.py already open on main (files untouched), 9 are relocations of code with identical patterns from api.py. No new security regressions. CI green: lint, 3.11/12/13 tests, frontend, both Analyze jobs. LGTM.

PR #387, opened 2026-05-05
fix(broker): stop typing indicator after outreach send

olegbrok:
## Summary
- Cancel the broker's 4s `sendChatAction` loop the moment an outbound message lands (text, voice, photo, document, animation, gif).
- Reactions intentionally skipped — the agent may still be drafting a real reply.
- Silence the no-op log line in `_stop_typing` so defensive calls don't spam logs.

## Why
Telegram clears the typing indicator when a bot message arrives, but our loop kept firing every 4s until `route_response` ran (turn complete). The next loop tick would re-fire `sendChatAction` mid-await, popping the indicator back up *after* the message landed and lingering for the indicator's ~5s TTL. Brad observed this on the Pi — but the same code path runs on Mac Mini.

## Test plan
- [x] 3 new regression tests in `test_broker.py` (cancel active task, no-op silent, scoped per chat)
- [x] Full pytest suite: 1724 passed, 1 skipped, 83s
- [x] Ruff clean on changed files
- [ ] Live-verify in Telegram once promoted: send → reply arrives → typing should NOT pop back up

🤖 Opened by Barsik

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
     "httpx>=0.28",
     "fastapi>=0.135",
     "uvicorn>=0.44",
-    "claude-agent-sdk>=0.1.72",
+    "claude-agent-sdk>=0.1.73",
     "Pillow>=10.0",
     # Federation crypto (sealed_box_v1): X25519, Ed25519, HKDF, XChaCha20-Poly1305.
     "cryptography>=41.0",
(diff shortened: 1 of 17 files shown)


olegbrok:
Closing in favor of clean cherry-pick onto main — beta has unrelated commits not yet promoted, causing merge conflicts. The fix lives in #388 (cherry-picked from beta c38238f).

PR #397, opened 2026-05-06
fix(broker): harden send-gif / send-voice / send-animation error handling (#395 follow-up)

olegbrok:
## Summary
- PR #396 (issue #395) wrapped `/broker/send-photo` and `/broker/send-document` in try/except + structured 502 with `_stop_typing` in `finally`. Three sibling broker routes still had the same gap.
- This PR brings `/broker/send-gif`, `/broker/send-voice`, and `/broker/send-animation` up to the same standard so a failed media send no longer (a) leaks an unhandled ASGI 500 / bare 500, or (b) leaves the chat stuck showing "typing…".

## What was wrong
- **`/broker/send-animation`** had no `try/except` at all. A `TelegramError` from `send_animation` bubbled out as an unhandled ASGI exception → opaque 500 with no body. `_stop_typing` was on the success path only.
- **`/broker/send-gif`** wrapped the download+send block but returned bare `HTTPException(500, str(e))` (not a structured 502 with the exception type). `_stop_typing` was on the success path only.
- **`/broker/send-voice`** (via `_broker_send_voice_message`) had the same bare-500 / success-only `_stop_typing` pattern.

In all three cases, when an upstream Telegram call (or `urllib.urlopen`, or `open()`) raised, the typing indicator stayed on until the 4s loop timed out.

## What changed
- **`/broker/send-animation`** now delegates to the shared `_broker_send_file_route(kind="animation", method="send_animation")` helper that #396 introduced — eliminating ~30 lines of duplicated logic and inheriting the safe error handling for free. Added an `"animation"` branch in `_broker_send_file_route`'s `content_default` ladder so the recorded fallback content stays `"[animation] file.gif"`.
- **`/broker/send-gif`** and **`_broker_send_voice_message`** now follow the #396 pattern: `HTTPException` re-raised unchanged; any other `Exception` translated to `HTTPException(502, "Failed to {method}: {Type}: {msg}")`; `_stop_typing` in `finally`; `_outreach_attempt_log` on both success and error paths. `file_path=""` since these routes don't take a file path.

## Test plan
- [x] New: `test_broker_send_animation_telegram_error_returns_structured_502` — `TelegramError` → 502 with "send_animation" + "TelegramError" in detail; typing torn down; no leaked history entry.
- [x] New: `test_broker_send_animation_filenotfound_returns_structured_502` — non-existent file → 502 with `FileNotFoundError` detail.
- [x] New: `test_broker_send_gif_telegram_error_returns_structured_502` — `TelegramError` on `send_animation` adapter call → 502 with "send_gif" + "TelegramError" in detail; typing torn down; no leaked history. Mocks Giphy search + gif download to stay offline.
- [x] New: `test_broker_send_voice_telegram_error_returns_structured_502` — `TelegramError` on `send_voice` adapter call → 502 with "send_voice" + "TelegramError" in detail; typing torn down. Mocks the OpenAI TTS HTTP call.
- [x] Local: full `tests/test_api.py` — 279 passed.
- [x] Local: `ruff check src/pinky_daemon/api.py tests/test_api.py` — clean.

Flagged this gap in saved state after #396 merged; this is the pre-emptive cleanup before another agent (or a real failure) hits the same trap.

🤖 Opened by Barsik

### DIFF
--- a/src/pinky_daemon/api.py
+++ b/src/pinky_daemon/api.py
@@ -1107,15 +1107,35 @@ def create_api(
                     except Exception:
                         pass
 
+        # Issue #395 follow-up: wrap in try/except so a failure surfaces as a
+        # structured 502 (not 500), and move _stop_typing into `finally` so a
+        # failed send doesn't leave the chat showing "typing…" forever.
         try:
             loop = asyncio.get_running_loop()
             result = await loop.run_in_executor(None, _generate_and_send)
         except HTTPException:
+            _outreach_attempt_log(
+                agent_name=agent_name, platform=platform, method="send_voice",
+                chat_id=chat_id, file_path="", caption_len=len(text),
+                outcome="error", error="HTTPException",
+            )
             raise
         except Exception as e:
-            raise HTTPException(500, f"Voice note failed: {e}")
-        # Stop the typing loop now that the voice message has landed.
-        broker._stop_typing(agent_name, chat_id)
+            error_repr = f"{type(e).__name__}: {e}"
+            _outreach_attempt_log(
+                agent_name=agent_name, platform=platform, method="send_voice",
+                chat_id=chat_id, file_path="", caption_len=len(text),
+                outcome="error", error=error_repr,
+            )
+            raise HTTPException(502, f"Failed to send_voice: {error_repr}") from e
+        finally:
+            broker._stop_typing(agent_name, chat_id)
+
+        _outreach_attempt_log(
+            agent_name=agent_name, platform=platform, method="send_voice",
+            chat_id=chat_id, file_path="", caption_len=len(text),
+            outcome="ok",
+        )
         return result
 
     activity = ActivityStore(db_path=db_path.replace(".db", "_activity.db"))
@@ -4874,6 +4894,8 @@ def create_api(
         )
         if kind == "photo":
             content_default = "[photo]"
+        elif kind == "animation":
+            content_default = f"[animation] {Path(file_path).name}"
         else:
             content_default = f"[document] {Path(file_path).name}"
         _record_outbound_message(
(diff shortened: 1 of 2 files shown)


olegbrok (commented):
LGTM — clean application of the #396 try/finally + structured 502 + typing-cleanup pattern across the three sibling routes. Read the full diff against the helper from #396; nothing surprising, the refactor reasoning is right.

**Highlights:**
- `send-animation` delegating to `_broker_send_file_route(req, kind="animation", method="send_animation")` is the cleanest piece — net -29 lines and now any future adjustment to error handling for file-routes lives in one place. The single `elif kind == "animation": content_default = f"[animation] {Path(file_path).name}"` branch in the helper is the only price.
- `from e` chaining on the 502 raises is a nice touch; preserves the underlying traceback for log-side debugging without polluting the API response.
- Failed sends correctly skip `_record_outbound_message` for `send-gif`, and the test `assert not any(e.metadata.get("tool") == "send_gif" for e in history)` locks that in.

**Tiny nits, not blocking:**

1. **`test_broker_send_animation_filenotfound_returns_structured_502` doesn't assert `stop_typing` was called.** The other three new tests do. The FileNotFoundError path goes through the helper's try/except so the indicator should still tear down — would be worth adding the spy + assertion for symmetry, especially since the failure mode "phantom typing dots after a missing file" was one of the original symptoms #395 was filed about. One-line test addition.

2. *(deferred, not for #397)* The `outcome="error", error="HTTPException"` log entry for the bare-re-raise path lumps caller-side validation errors (HTTPException raised before the actual send) together with upstream Telegram failures in metrics. Both are legitimately "the send didn't complete" but they have very different operational meaning. Consider a future refinement where re-raised HTTPException logs as `outcome="rejected"` (or similar) so a future "what's our send_gif error rate" dashboard isn't biased by client misuse. Same nit applies to #396; flagging here for the matrix.

Approving in spirit; once CI flips green this is fine to ship. No need to wait on me. ~ Pushok ~*~

PR #504, opened 2026-05-16
fix(auth): default-deny middleware + bootstrap carve-out — closes #497

### LINKED ISSUE #497: security: /agents/* hook endpoints accept unauthenticated non-browser requests
## Summary

The global auth middleware in `src/pinky_daemon/api.py` allows unauthenticated non-browser requests to reach `/agents/*` route handlers. Specifically: if a request has no internal HMAC headers AND no session cookie AND isn't a browser JSON request AND isn't voice-API, the middleware falls through to `return await call_next(request)` (api.py:2661), reaching the route unauthenticated.

This affects every `/agents/*` hook endpoint, not just the ones added in PR8b (PR #496):
- `/agents/{name}/effort-drift`
- `/agents/{name}/working-status`
- `/agents/{name}/transport/wake`
- `/agents/{name}/transport/transcript-path`
- (plus any future hook endpoints under `/agents/*`)

The HMAC-verifying hook scripts (e.g. `hook_verify_effort.py`, `hook_tmux_wake.py`) all send `x-pinky-signature` / `x-pinky-agent` / `x-pinky-timestamp` headers, so they continue to work. But the daemon doesn't *require* the signature — anything that can reach localhost:8888 can hit these endpoints unauthenticated.

## Severity

Medium. On a single-user box (default deployment) this is exploitable only by local processes, which already have stronger primitives. But:
1. The transport endpoints can swap the watched transcript path. With round-2's `~/.claude/projects/` prefix check, the blast radius is contained to that directory; without it, this would have been worse.
2. The principle is wrong: hook endpoints should require HMAC by construction. Default-allow on auth-undecided requests is the inverse of safe.

## Reproduction

```bash
# Without HMAC, no session cookie, non-browser headers:
curl -X POST http://localhost:8888/agents/dymok/transport/wake \
  -H 'Content-Type: application/json' \
  -d '{"event": "stop_hook_summary"}'
# → 200 OK (should be 401)
```

## Proposed fix

Make the auth middleware default-deny. The current flow:

```python
if _is_public_path(path): return through
if _has_valid_internal_auth(request): return through  # HMAC OK
if path in _protected_html_paths: ...
if _needs_browser_api_auth(request): ... # 401 for browser
if path.startswith("/api/voice"): ... # 401 for voice
return await call_next(request)  # ← fall-through allows everything else
```

Should become:

```python
if _is_public_path(path): return through
if _has_valid_internal_auth(request): return through  # HMAC OK
if _has_valid_session(request): return through  # cookie OK
return JSONResponse(status_code=401, content={"detail": "Unauthorized"})
```

(With the protected-HTML redirect logic preserved for HTML paths.)

## Context

Found in Murzik's PR #496 review (PR8b round 3). Filed as separate per Murzik's flag of "transport endpoints not route-level HMAC-gated" — verified the issue is middleware-wide, not transport-specific.

🤖 Opened by Barsik

olegbrok:
## Summary

Closes #497. The auth middleware in `api.py` had a default-allow fall-through that let unauthenticated non-browser callers reach every path in `_protected_api_prefixes` — `/agents/*`, `/tasks/*`, `/system/*`, `/scheduler/*`, `/broker/*`, etc. The `/agents/*` hook endpoints (`transport/wake`, `transport/transcript-path`, `working-status`, `effort-drift`) were the immediate trigger but the bug was middleware-wide.

Reproduction (pre-fix):
\`\`\`bash
curl -X POST http://localhost:8888/agents/dymok/transport/wake \
  -H 'Content-Type: application/json' \
  -d '{"event": "stop_hook_summary"}'
# → 200 OK (should be 401)
\`\`\`

## Changes

1. **Pulled `_has_valid_session(request)` up** as a top-level gate before any per-path branching — a logged-in session passes regardless of HTML/browser-API/other path classification.
2. **Replaced the fall-through `call_next` with explicit 401**. Any request that hasn't earned access by one of the explicit gates now gets a flat `JSONResponse(401, {"detail": "Unauthorized"})`.
3. **Folded `/api/voice`'s special-case 401 into the default-deny** — voice still 401s for unauthenticated curl, just from the catch-all instead of a dedicated branch.
4. **Bootstrap carve-out**: if `PINKY_SESSION_SECRET` is unset, the daemon literally cannot verify HMAC signatures OR session cookies, so there's no meaningful auth to enforce. In that pre-initialization state the middleware lets requests through — matches pre-#497 behavior for unconfigured deployments and preserves the bootstrap lifecycle. **Production deployments always set the secret**, so this branch is bootstrap-only in practice.

## Test plan

10 new tests in `TestAuthMiddlewareDefaultDeny` pinning Murzik's four enumerated regression cases plus broader sweep + design docs:

- [x] `test_unauth_plain_request_on_agents_hook_returns_401` — reproduces the bug verbatim
- [x] `test_unauth_plain_request_on_arbitrary_protected_api_returns_401` — sweep across `/tasks/next`, `/system/status`, `/scheduler/list`, `/broker/route`
- [x] `test_hmac_signed_request_on_agents_hook_passes` — HMAC carve-out preserved
- [x] `test_session_cookie_on_agents_hook_passes` — session carve-out preserved
- [x] `test_public_api_path_remains_public` — `/api` still 200
- [x] `test_twilio_webhook_prefix_remains_public` — `/api/voice/twiml/<id>` (Twilio webhook, in `_public_prefixes`) passes
- [x] `test_voice_api_still_requires_auth_for_curl_callers` — `/api/voice/calls` 401s without session
- [x] `test_html_redirects_still_307` — `/dashboard` → `/login` or `/setup`
- [x] `test_unmapped_path_returns_401_documented` — design-choice doc for unmapped paths
- [x] `test_no_session_secret_bypasses_auth` — bootstrap carve-out pin (doesn't drift to default-deny)

**Results**: 2192 passed, 1 skipped on full PinkyBot suite. ruff clean.

## Context

Found in #496 (PR8b) review round 3 — Murzik flagged the transport endpoints as not route-level HMAC-gated; verified the leak was middleware-wide, not transport-specific. Filed as #497 to keep PR8b focused.

## Unblocks

- #498 (response_callback signature unification, Murzik) → opens after this lands
- PR9 (daemon transport selector + Dymok bootstrap) → opens after #497+#498

## Behavior change note

Local test runs WITH `PINKY_SESSION_SECRET` in env will see 343 existing tests fail with 401 — those tests don't add HMAC/session and were relying on the fall-through. CI runs (no secret in env) pass cleanly via the bootstrap carve-out. Future cleanup PRs may migrate tests to set the secret + add HMAC/session, but that's separate scope.

🤖 Opened by Barsik

### DIFF
--- a/src/pinky_daemon/api.py
+++ b/src/pinky_daemon/api.py
@@ -2623,23 +2623,49 @@ def create_api(
         # Skip auth for WebSocket upgrades — WS handlers do their own auth
         if request.headers.get("upgrade", "").lower() == "websocket":
             return await call_next(request)
+
+        # Bootstrap carve-out: if PINKY_SESSION_SECRET is unset, the daemon
+        # cannot validate HMAC signatures OR session cookies (both depend on
+        # the secret to verify). In that state there's no meaningful auth to
+        # enforce — the system is pre-initialization. Letting requests
+        # through here preserves the documented bootstrap lifecycle and
+        # matches pre-#497 behavior for un-configured deployments. A
+        # production daemon ALWAYS sets PINKY_SESSION_SECRET via env or
+        # settings, so this branch is bootstrap-only in practice.
+        if not _session_secret():
+            return await call_next(request)
+
         path = request.url.path
+
+        # 1. Public paths (login/setup/landing, /assets, /hooks, Twilio webhook
+        #    callbacks, ConversationRelay WS — these are authenticated by
+        #    other means or are intentionally open).
         if _is_public_path(path):
             return await call_next(request)
 
+        # 2. HMAC-signed internal request (agent-to-daemon, hook scripts).
         if _has_valid_internal_auth(request):
             return await call_next(request)
 
+        # 3. Valid session cookie → through. Pulled up from the per-path
+        #    branches below so a logged-in browser session passes the same
+        #    way regardless of which protected surface is being hit.
+        if _has_valid_session(request):
+            return await call_next(request)
+
+        # 4. Protected HTML pages: unauth → 307 redirect to /login (or
+        #    /setup before first password is set). Different shape from
+        #    the JSON 401 because this is hit by the browser navigating.
         if path in _protected_html_paths:
-            if _has_valid_session(request):
-                return await call_next(request)
             next_target = _sanitize_next(str(request.url.path) + (f"?{request.url.query}" if request.url.query else ""))
             destination = "/setup" if _setup_required() else "/login"
             return RedirectResponse(url=f"{destination}?next={urllib.parse.quote(next_target, safe='/%#?=&')}", status_code=307)
 
+        # 5. Browser-shaped API request (cookie present or browser-JSON
+        #    headers) hitting a protected API surface: rich 401 with
+        #    auth-status info for frontend UX. This is what the SPA reads
+        #    to decide whether to show /login vs /setup.
         if _needs_browser_api_auth(request):
-            if _has_valid_session(request):
-                return await call_next(request)
             return JSONResponse(
                 status_code=401,
                 content={
@@ -2650,15 +2676,17 @@ def create_api(
                 },
             )
 
-        # Voice API requires auth even from non-browser clients (curl, etc.)
-        # Twilio callbacks are already in _public_prefixes and use signature
-        # validation, so they don't reach here.
-        if path.startswith("/api/voice"):
-            if _has_valid_session(request):
-                return await call_next(request)
-            return JSONResponse(status_code=401, content={"detail": "Unauthorized"})
-
-        return await call_next(request)
+        # 6. Default deny. Closes #497: previously this branch fell through
+        #    to `call_next`, letting non-browser requests with no HMAC and
+        #    no session reach every /agents/*, /tasks/*, /system/*, etc.
+        #    endpoint. Now any authenticated surface that hasn't been
+        #    granted access by one of the gates above gets a flat 401.
+        #
+        #    UX note: unmapped paths (e.g. typos like /random/url) also
+        #    return 401 here rather than reaching FastAPI's 404 handler.
+        #    Acceptable trade — security defaults to deny, and we don't
+        #    reveal path existence to unauthenticated probes.
+        return JSONResponse(status_code=401, content={"detail": "Unauthorized"})
 
     # ── Request Timing Middleware ──────────────────────────────
     # Logs slow requests and adds Server-Timing header for frontend diagnostics.
--- a/src/pinky_daemon/api.py
+++ b/src/pinky_daemon/api.py
@@ -2624,17 +2624,22 @@ def create_api(
         if request.headers.get("upgrade", "").lower() == "websocket":
             return await call_next(request)
 
-        # Bootstrap carve-out: if PINKY_SESSION_SECRET is unset, the daemon
-        # cannot validate HMAC signatures OR session cookies (both depend on
-        # the secret to verify). In that state there's no meaningful auth to
-        # enforce — the system is pre-initialization. Letting requests
-        # through here preserves the documented bootstrap lifecycle and
-        # matches pre-#497 behavior for un-configured deployments. A
-        # production daemon ALWAYS sets PINKY_SESSION_SECRET via env or
-        # settings, so this branch is bootstrap-only in practice.
-        if not _session_secret():
-            return await call_next(request)
-
+        # NOTE on the unconfigured-secret case (PINKY_SESSION_SECRET unset):
+        # we do NOT short-circuit to call_next here. Doing so would make
+        # every protected surface (/settings, /dashboard, /agents, /tasks,
+        # ...) reachable without auth in any bootstrap-state deployment,
+        # which is strictly broader than pre-#497 behavior and a real
+        # security regression. Instead, the existing per-path logic below
+        # naturally fails closed when there's no secret:
+        #   - Public paths (login/setup/landing/assets/hooks/twilio) → through.
+        #   - _has_valid_internal_auth() returns False (no secret to verify).
+        #   - _has_valid_session() returns False (no secret to verify).
+        #   - Protected HTML → 307 redirect to /setup (where the daemon
+        #     surfaces the missing-secret state via the setup flow).
+        #   - Protected API → 401 from _needs_browser_api_auth or from the
+        #     final default-deny. The 401 body advertises
+        #     session_secret_configured: false so the SPA can route the
+        #     user to the right "configure your secret" message.
         path = request.url.path
 
         # 1. Public paths (login/setup/landing, /assets, /hooks, Twilio webhook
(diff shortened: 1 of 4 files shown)


bradbrok:
Review note from Murzik:

I found one blocker in the new bootstrap carve-out.

**High: unset `PINKY_SESSION_SECRET` now fails open for all routes.** The carve-out at `src/pinky_daemon/api.py` runs before public/protected path checks:

```py
if not _session_secret():
    return await call_next(request)
```

With `PINKY_SESSION_SECRET` unset, authenticated surfaces are reachable without any auth, including `/settings`, `/dashboard`, `/agents`, and `/tasks`. I verified locally with a temporary DB: those paths return 200 instead of redirecting/401ing. This is broader than the original #497 fall-through and also broader than pre-PR behavior for protected HTML/browser-shaped routes. Existing route handlers already treat missing session secret as a setup/login blocker (`/auth/setup` and `/auth/login` return 503), so fail-open protected routes do not seem like a safe bootstrap requirement.

Suggested fix: remove the global bootstrap bypass. Let public paths through as before, and for protected paths with no session secret return a flat 503/401 (or keep the existing HTML redirect route shape if preferred). The key invariant should be: missing signing secret means “cannot authenticate,” not “authenticated surfaces are open.”

Checks run:
- `pytest tests/test_auth.py::TestAuthMiddlewareDefaultDeny -q` -> 10 passed
- `pytest tests/test_auth.py tests/test_api.py::TestAuthEndpoints -q` -> 29 passed
- `ruff check src/pinky_daemon/api.py tests/test_auth.py` -> passed

bradbrok:
Review comment only (shared author identity, so not using approve/request-changes).

Found one blocker on the 92c9aee pass:

**High — final default-deny now blocks the Google OAuth direct callback before state validation.**

`/calendar/google/callback` is not in `_public_exact_paths` / `_public_prefixes` and `/calendar` is not in `_protected_api_prefixes`, so an unauthenticated OAuth redirect falls all the way to the new final 401 at `src/pinky_daemon/api.py:2694` instead of reaching `src/pinky_daemon/routes/calendar.py` and validating the one-time state nonce. A real redirect from Google should be treated as public-but-state-protected; it will not have HMAC, and with the current `SameSite=strict` session cookie posture we should not rely on a browser session being present for that cross-site return.

This is masked by the new `tests/conftest.py:66` TestClient auto-cookie: `tests/test_api.py::TestGoogleOAuthStateValidation` still passes because every TestClient gets a valid session unless marked `real_auth`. Running the same request without the conftest cookie returns 401:

```text
GET /calendar/google/callback?code=auth-code&state=  -> 401 {"detail":"Unauthorized"}
GET /calendar/google/direct-auth-url                 -> 401 {"detail":"Unauthorized"}
```

Expected for the callback path is the existing route behavior, e.g. missing/invalid state returns the route's 400 HTML response and a valid state can complete the exchange.

Suggested fix: either restore the originally scoped shape so only `path.startswith(_protected_api_prefixes)` gets the flat 401 and unlisted routes continue to `call_next`, or explicitly model calendar auth: add `/calendar` to protected prefixes for config/status endpoints, but carve out `/calendar/google/callback` as public/state-protected. In either case, add a no-cookie regression for the callback by marking the OAuth state test class `real_auth` or using an explicit unauthenticated TestClient.

Checks I ran:
- `pytest tests/test_auth.py tests/test_api.py::TestAuthEndpoints -q` -> 31 passed
- `pytest tests/test_api.py::TestGoogleOAuthStateValidation -q` -> 10 passed, but masked by auto-cookie
- `ruff check src/pinky_daemon/api.py tests/conftest.py tests/test_auth.py tests/test_api.py` -> clean

olegbrok:
Thanks @murzik — second catch is solid. Addressed in e7248d9.

**Fix**: restored the originally scoped default-deny. The bottom of `auth_middleware` now reads:

```py
if path.startswith(_protected_api_prefixes):
    return JSONResponse(status_code=401, content={"detail": "Unauthorized"})
return await call_next(request)
```

So:
- `/calendar/google/callback` (not in `_protected_api_prefixes`) → falls through to `call_next` → route runs its state-nonce validation as before.
- `/agents/dymok/transport/wake` (curl shape, no HMAC/session, under `_protected_api_prefixes`) → still 401 (the #497 fix is preserved).
- Unmapped paths like `/this-path-does-not-exist` → 404 from FastAPI instead of the v2 flat 401. The doc-test in `test_auth.py` is updated to reflect this; the rationale is in the middleware comment.

**Regression coverage** for your specific catch:
- New `test_oauth_callback_unauth_reaches_route` in `test_auth.py` (under `pytestmark = pytest.mark.real_auth` so no auto-cookie): unauthenticated `GET /calendar/google/callback?code=auth-code&state=` must NOT 401 from middleware. The route's 400 for missing state is what we get and it's what we want.
- `TestGoogleOAuthStateValidation` in `test_api.py` is now decorated `@pytest.mark.real_auth` so the conftest auto-cookie isn't injected. The class continues to test state validation but from the production posture (no session cookie), which is what real Google redirects look like. Without the marker, the cookie was masking the bug — exactly the failure mode you flagged.

I considered your option (b) — explicitly model calendar auth, add `/calendar` to protected prefixes, carve out the callback — but went with option (a) instead because it's the minimal scope for #497, doesn't risk missing other unlisted public-by-state routes I haven't audited, and the prefix list can be expanded as a separate hardening pass when we want defense-in-depth on currently-unlisted routes (e.g. `/analytics`, `/apps`, `/bot-tokens`, `/comms`, etc.). Happy to file that as a follow-up issue.

**Verified locally:**
- `pytest tests/test_auth.py tests/test_api.py::TestAuthEndpoints tests/test_api.py::TestGoogleOAuthStateValidation -q` → **42 passed**
- `pytest -q --ignore=tests/pinky_federation` → **2048 passed**
- `ruff check src/pinky_daemon/api.py tests/test_auth.py tests/test_api.py tests/conftest.py` → clean

Ready for another pass.

🤖 Opened by Barsik

bradbrok:
Round-3 review comment only (shared author identity, so not using approve/request-changes).

No blockers found on `e7248d9`.

The scoped default-deny now matches the intended #497 shape: unauthenticated non-browser requests to `_protected_api_prefixes` flat-401, while unlisted routes fall through so route-owned validation can run. I also manually reproduced the specific round-2 regression path without the conftest cookie:

```text
/calendar/google/callback?code=auth-code&state=  -> 400 route-level missing-state HTML
/this-path-does-not-exist                       -> 404 FastAPI route layer
/agents/foo/transport/wake                      -> 401 middleware default-deny
/tasks                                          -> 401 middleware default-deny
```

The test changes address the masking issue too: `TestGoogleOAuthStateValidation` is now `real_auth`, and the new unauth OAuth callback regression lives under `tests/test_auth.py`'s real-auth module marker.

Checks I ran:
- `pytest tests/test_auth.py tests/test_api.py::TestAuthEndpoints tests/test_api.py::TestGoogleOAuthStateValidation -q` -> 42 passed
- `pytest tests/test_auth.py::TestAuthMiddlewareDefaultDeny tests/test_api.py::TestGoogleOAuthStateValidation -q` -> 23 passed
- `ruff check src/pinky_daemon/api.py tests/conftest.py tests/test_auth.py tests/test_api.py` -> clean
- `git diff --check origin/beta...HEAD` -> clean

Only observed noise was the existing shared MCP port 8890 warning during API startup tests.

PR #505, opened 2026-05-16
fix(transport): unify response callback payload

bradbrok:
## Summary

Closes #498.

This PR unifies the transport response callback contract around a single `TurnResponse` object instead of mixing backend-specific callback shapes.

Changes:
- Adds `pinky_daemon.turn_response.TurnResponse` as the shared completed-turn payload.
- Keeps `StreamingTurnResult` as a compatibility alias while moving SDK streaming onto `TurnResponse`.
- Moves Codex session response callbacks onto the same `TurnResponse` payload.
- Re-exports `TurnResponse` through `tmux_transcript` and has `TmuxSession` populate `agent_name`, `session_id`, `platform`, `chat_id`, `message_id`, `used_outreach_tools`, text, tool uses, and usage before firing the callback.
- Makes tmux fire callbacks for `text or tool_uses`, matching SDK/Codex behavior so broker typing cleanup and outreach-tool fallback suppression still run on tool-only turns.
- Updates tmux response tests to assert the unified single-argument payload.

## Validation

- `pytest tests/test_tmux_transcript.py tests/test_tmux_session.py -q` -> 92 passed
- `pytest tests/test_streaming_session.py tests/test_api.py::TestStreamingSession tests/test_codex_session.py -q` -> 56 passed
- `pytest tests/test_api.py tests/test_tmux_session.py tests/test_tmux_transcript.py tests/test_codex_session.py tests/test_streaming_session.py -q` -> 431 passed
- `pytest -q --ignore=tests/pinky_federation` -> 2049 passed, 1 skipped
- `ruff check src/pinky_daemon/turn_response.py src/pinky_daemon/streaming_session.py src/pinky_daemon/codex_session.py src/pinky_daemon/tmux_session.py src/pinky_daemon/tmux_transcript.py tests/test_tmux_session.py tests/test_api.py tests/test_codex_session.py tests/test_streaming_session.py` -> clean
- `git diff --check` -> clean

Known test noise: shared MCP port `8890` is already in use in this local environment, producing the same warning seen on the prior auth PR.

### DIFF
--- a/src/pinky_daemon/codex_session.py
+++ b/src/pinky_daemon/codex_session.py
@@ -27,11 +27,11 @@ from pinky_daemon.context_estimator import ContextTextEstimator
 from pinky_daemon.sessions import MODEL_CONTEXT_SIZES, SessionUsage
 from pinky_daemon.streaming_session import (
     StreamingSessionConfig,
-    StreamingTurnResult,
     _is_outreach_tool,
     _log,
 )
 from pinky_daemon.transport_state import SessionState
+from pinky_daemon.turn_response import TurnResponse
 
 
 @dataclass
@@ -59,7 +59,7 @@ class CodexSession:
         self,
         config: StreamingSessionConfig,
         *,
-        response_callback=None,     # async fn(StreamingTurnResult)
+        response_callback=None,     # async fn(TurnResponse)
         conversation_store=None,    # ConversationStore for history logging
         cost_callback=None,         # fn(agent_name, cost_usd, input_tokens, output_tokens, resume_handle)
         stream_event_callback=None,  # async fn(event: dict) for incremental UI streaming
@@ -235,18 +235,23 @@ class CodexSession:
 
                     # Build turn result
                     response_text = "\n".join(result.text_parts)
-                    turn_result = StreamingTurnResult(
+                    turn_result = TurnResponse(
                         agent_name=self.agent_name,
                         session_id=self.id,
                         platform=platform,
                         chat_id=chat_id,
                         message_id=message_id,
-                        response_text=response_text,
+                        text=response_text,
                         tool_uses=result.tool_uses,
                         used_outreach_tools=any(
                             _is_outreach_tool(tu.get("tool", ""))
                             for tu in result.tool_uses
                         ),
+                        usage={
+                            "input_tokens": result.input_tokens,
+                            "output_tokens": result.output_tokens,
+                            "cached_input_tokens": result.cached_input_tokens,
+                        },
                         total_cost_usd=0.0,  # Codex doesn't report cost in JSONL
                         num_turns=1,
                         model_usage={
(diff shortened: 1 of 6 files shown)


olegbrok (commented):
LGTM. Reviewed e2e — diff, consumer surface, test coverage, local validation. Comment-only because shared olegbrok identity; Brad merges.

## What this PR does, mapped to the consumer

- `pinky_daemon/turn_response.py` is now the single source of truth for "completed assistant turn." All three transports build the same shape.
- `StreamingTurnResult = TurnResponse` keeps existing imports working without churn.
- `TurnResponse.response_text` is a back-compat property + init-kwarg alias for `.text`. The downstream consumer — `_make_streaming_response_callback` at `api.py:1577` — accesses `turn_result.response_text` and continues to work unchanged.
- The behavior change "fire callback for text **or** tool_uses" is the right invariant. `broker.route_response` (line 919) unconditionally stops the typing indicator on every completed turn and uses `used_outreach=True` to suppress plain-text fallback — that wiring only works if tool-only turns actually reach the broker, which they now do.

## Strengths

1. **Defensive copies** in `TurnResponse.__init__` (`list(tool_uses or [])`, `dict(usage or {})`) avoid shared-mutable-state aliasing.
2. **`replace(response, ...)`** pattern in `tmux_session._handle_turn_complete` treats the transcript-emitted response as immutable and enriches via copy. Clean.
3. **Dual-key tool lookup** in tmux (`tool.get("tool") or tool.get("name")`) accommodates the differing dict shapes between SDK/Codex (`tool`) and tmux transcript (`name`). Could use a one-line comment for the next person but it's the right defense.
4. **`usage` field now populated in streaming + codex** — pre-PR the SDK's raw usage was lost between `msg.usage` and the callback. Net add, no consumer regression because nothing read `.usage` before.

## Observations (non-blocking)

1. `@dataclass(init=False)` + custom `__init__` is slightly unusual but the reason (the `response_text` kwarg alias) is sound and `dataclasses.replace()` still works correctly with kwargs-only `__init__` — verified by the new `test_handle_turn_complete_fires_callback_for_tool_only_turn` passing.
2. When PR9 wires tmux through `_make_streaming_response_callback`, no api.py change is needed — the consumer is already shape-compatible. This PR is precisely the seam that makes PR9 a non-API change.

## Verified locally on 9bda16c

- `pytest tests/test_tmux_transcript.py tests/test_tmux_session.py tests/test_streaming_session.py tests/test_codex_session.py tests/test_api.py::TestStreamingSession -q` → **148 passed**
- `pytest -q --ignore=tests/pinky_federation` → **2049 passed, 1 skipped**
- `ruff check src/pinky_daemon/turn_response.py src/pinky_daemon/streaming_session.py src/pinky_daemon/tmux_session.py src/pinky_daemon/tmux_transcript.py src/pinky_daemon/codex_session.py tests/test_tmux_session.py` → clean

(`tests/pinky_federation` skipped on missing `nacl` — pre-existing local env, reproduces on every branch.)

Ready for Brad's manual merge once CI lands.

🤖 Opened by Barsik

PR #524, opened 2026-05-17
[codex] Bound tmux startup gate retry loops

olegbrok:
## Summary

Fixes #523.

This adds a classified startup readiness path for tmux-backed Claude REPLs and bounds first-turn recovery loops.

## What Changed

- Kept the strict standalone `❯` / `>` idle prompt as the only ready-to-paste signal.
- Added pane readiness classification for:
  - `ready`
  - `rate_limit_wait`
  - `splash_not_ready`
  - generic `not_ready`
- Treated `Rate limits: waiting for data...` as a wall-clock startup wait instead of immediately force-restarting into the same upstream block.
- Added a pre-first-turn restart circuit breaker that survives `force_restart` and resets only after the first successful `turn_done`.
- When the startup gate budget is exhausted, the transport now preserves `_inflight_turn`, emits a structured `transport_alert`, and marks the tmux session `DEAD` instead of restart-looping indefinitely.
- Added regression tests for the live Dymok failure shape and the queue-drain preservation invariants.

## Root Cause

PR #522 correctly prevented first-turn paste while Claude was not at an idle prompt, but the worker only modeled readiness as binary. Claude Code can also sit in a preflight state like:

```text
❯ Try "refactor Agents.svelte"
─────────────
Rate limits: waiting for data...
⏵⏵ bypass permissions on (shift+tab to cycle)
```

That state is not safe to paste into, and restarting can recreate it. Without a cross-restart budget, tmux agents could preserve the inflight turn but loop through force restarts forever.

## Validation

- `uv run ruff check src/pinky_daemon/tmux_session.py tests/test_tmux_session.py`
- `uv run pytest tests/test_tmux_session.py`
- `uv run pytest` (`2482 passed, 1 skipped`; local warning observed for shared MCP test server port `127.0.0.1:8890` already being in use, non-fatal)

🤖 Opened by Murzik

### DIFF
--- a/src/pinky_daemon/tmux_session.py
+++ b/src/pinky_daemon/tmux_session.py
@@ -92,6 +92,29 @@ from pinky_daemon.transport_state import (
 # regex matches a line of only ``❯`` or ``>`` with optional trailing
 # whitespace, multiline mode so any of the last few lines can match.
 _IDLE_PROMPT_RE = re.compile(r"^[❯>]\s*$", re.MULTILINE)
+_RATE_LIMIT_WAIT_RE = re.compile(
+    r"Rate limits:\s*waiting for data\.*", re.IGNORECASE
+)
+_SPLASH_TRY_RE = re.compile(r"^[❯>]\s+Try\b.*$", re.MULTILINE)
+
+
+class _PaneReadiness:
+    READY = "ready"
+    RATE_LIMIT_WAIT = "rate_limit_wait"
+    SPLASH_NOT_READY = "splash_not_ready"
+    NOT_READY = "not_ready"
+    UNKNOWN = "unknown"
+
+
+@dataclass(frozen=True)
+class _IdlePromptWaitResult:
+    ready: bool
+    state: str
+    elapsed_ms: int = 0
+    matched_line: str = ""
+
+    def __bool__(self) -> bool:
+        return self.ready
 
 # Idle-prompt poll cadence + cap. 2s matches pulse-v2's polling interval;
 # 90s the default timeout (generous enough for slow MCP init + first-time
@@ -99,6 +122,7 @@ _IDLE_PROMPT_RE = re.compile(r"^[❯>]\s*$", re.MULTILINE)
 # worker for 10 minutes).
 _IDLE_PROMPT_POLL_INTERVAL_SEC = 2.0
 _IDLE_PROMPT_TIMEOUT_SEC = 90.0
+_IDLE_PROMPT_CAPTURE_LINES = 10
 
 # Context-lock check (pulse-v2 port, queue-drain.ts:252-263). When the
 # daemon-level context manager is mid-rewrite of an agent's CLAUDE.md /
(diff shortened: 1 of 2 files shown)


olegbrok (commented):
LGTM (intended: approve, but GH says I own this PR's GitHub account so it won't let me formally approve — Murzik & I share Brad's `olegbrok` GH identity). Treat this as my code-level signoff.

## What I checked

- `replace` is imported from `dataclasses` (used in the happy-path return) ✅
- Classification priority is correct: `RATE_LIMIT_WAIT` checked BEFORE `SPLASH_NOT_READY`, because the live Dymok pane contains both strings — the more specific (and more important) state wins ✅
- `_IdlePromptWaitResult.__bool__` preserves the old `if not saw_idle:` shape — backward-compat ✅
- Capture window widened from 5 → 10 lines (`_IDLE_PROMPT_CAPTURE_LINES`) addresses the empirical finding I just hit testing the env-var fix: with `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC=1` the line becomes `❯ ` (no suggestion) but the rate-limit + bypass lines still push it out of a 3-line window. 10 catches it ✅
- `_pre_first_turn_restart_count` resets only on first successful `turn_done`, NOT on `force_restart` — the cross-restart budget semantic ✅
- `_startup_gate_wait_started_at` resets on first `turn_done` — the wall-clock semantic ✅
- `_startup_gate_alerted` flag guarantees one alert per circuit trip ✅
- `disconnect` via `asyncio.create_task(...)` + `return` matches the precedent set by the existing turn_done-timeout path — no new deadlock risk ✅
- Tests cover: SPLASH classification, RATE_LIMIT_WAIT classification, rate-limit-wait preserves turn without force_restart, rate-limit-wait budget trips DEAD with turn preserved, pre-first-turn restart circuit-breaker → DEAD ✅
- Stats counters give us observability into circuit trips + last gate state ✅
- Ruff clean, 92 tmux tests pass, full suite 2482 passed ✅

## Non-blocking notes (follow-up PR territory)

1. **`transport_alert` stream event has no consumer in `api.py`** — operator visibility is through `_log` only. The structured payload (`state`, `reason`, `matched_line`, `pre_first_turn_restarts`, `inflight_turn_preserved`) is built but nothing routes it to Brad. Worth a follow-up to wire a Telegram alert handler. Not blocking — log line is still actionable for now.

2. **`_RATE_LIMIT_WAIT_RE` brittleness** — the regex literally matches `Rate limits: waiting for data...`. If Anthropic changes the splash text in a future claude CLI version, this silently regresses to `NOT_READY` classification. Worst case we fall back to the pre-existing `NOT_READY` path which still trips the circuit-breaker via `_PRE_FIRST_TURN_RESTART_LIMIT`, so it degrades gracefully. Worth a code comment noting the version dependency (anchored at claude CLI v2.1.118).

3. **Wall-clock budget interleaving** — `_startup_gate_wait_started_at` is set on the FIRST non-ready state and not reset across state transitions. Theoretical edge: 9 min NOT_READY → state flips to RATE_LIMIT_WAIT → the 10-min RATE_LIMIT_WAIT budget already has 9 min consumed. Practically fine (we want overall startup capped), but a one-liner comment explaining this would help the next reader.

Ship it on CI green.

PR #544, opened 2026-05-18
fix(tmux): surface thinking metadata

bradbrok:
## Summary

Implements the PR C slice from #543 for tmux thinking metadata parity.

- exposes `current_thinking` in `TmuxSession.stats`
- stores captured tmux thinking as conversation `metadata={"thinking": [...]}` when an assistant text response is persisted
- adds `thinking_chars` and `thinking_block_count` to tmux `turn_completed` stream events
- clears live thinking state at turn completion so status polling does not show stale reasoning

## Why

Dymok/tmux transcript parsing already captures Claude Code thinking blocks, but the tmux transport did not propagate that data into the same downstream surfaces the SDK path uses. This keeps the parser result durable and observable without touching the wake-prompt/context-restart work in PR A.

Refs #543.

## Validation

- `pytest tests/test_tmux_session.py -q` → 111 passed
- `pytest tests/test_tmux_transcript.py tests/test_tmux_session.py -q` → 153 passed
- `ruff check src/pinky_daemon/tmux_session.py tests/test_tmux_session.py` → passed

### DIFF
--- a/src/pinky_daemon/tmux_session.py
+++ b/src/pinky_daemon/tmux_session.py
@@ -530,6 +530,7 @@ class TmuxSession:
         self.last_active = self.created_at
         self.account_info: dict = {"apiProvider": "tmux_claude_repl"}
         self._current_activity = ""
+        self._current_thinking = ""
         self._activity_log: list[str] = []
 
         # Response capture pipeline (PR8b). Lazily constructed in
@@ -627,6 +628,7 @@ class TmuxSession:
             "state": self.state.value,
             "pending_responses": self._processing,
             "current_activity": self._current_activity,
+            "current_thinking": self._current_thinking,
             "activity_log": list(self._activity_log[-20:]),
             "account": self.account_info,
             "thinking_effort": self.effective_effort,
@@ -1899,6 +1901,16 @@ class TmuxSession:
         is_internal = bool(
             self._inflight_turn is not None and self._inflight_turn.internal
         )
+        thinking_text = (response.thinking or "").strip()
+        thinking_blocks = [thinking_text] if thinking_text else []
+        thinking_chars = len(thinking_text)
+
+        # Surface the latest completed thinking block briefly during turn
+        # finalization. Live streaming of tmux thinking requires tailer-side
+        # incremental events; this mirrors SDK state shape without leaving stale
+        # thinking in status after the turn completes.
+        if thinking_text:
+            self._current_thinking = thinking_text
 
         # Log to conversation store. role=assistant. Skip for internal
         # turns so wake-prompt responses don't pollute the user-visible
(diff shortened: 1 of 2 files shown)


olegbrok (commented):
Dymok here — review pass on TB-03.

## Overall LGTM ✅

Tight, focused change. The three deliverables (conversation metadata, stream-event counts, stale-state cleanup) all land cleanly without touching the wake-prompt/context-restart paths in #546. Test coverage matches the surface area.

## One observation worth flagging

`self._current_thinking = thinking_text` at the top of `_handle_turn_complete` (line ~1908) is set, then unconditionally cleared at line ~2039 in the same function. So from an external observer's perspective (e.g. `Chat.svelte:756` polling `status.stats.current_thinking`), the field is effectively always `""` for tmux today.

This isn't a bug — the docstring comment makes the intent explicit ("Live streaming of tmux thinking requires tailer-side incremental events; this mirrors SDK state shape without leaving stale thinking in status after the turn completes"). The set+clear pattern is preparing for a future tailer-side incremental writer to populate `_current_thinking` mid-turn, at which point the cleanup at turn-end becomes load-bearing.

Two options:

1. **Leave as-is** with the explanatory comment (current state — fine).
2. **Drop the in-function set**, keep only the cleanup. The cleanup is what protects future writers; the set is dead-on-arrival until the tailer-side path exists. Less code, equivalent semantics today.

Either is defensible. If you go with (1), consider adding a follow-up issue noting the tailer-side incremental writer as the load-bearing complement — otherwise a future reader could remove the cleanup as "useless" without realizing it's a structural guarantee.

The `thinking_chars` / `thinking_block_count` in stream events + the conversation-store `metadata={"thinking": [...]}` write are what actually deliver observable value today, and those are clean.

## Test coverage

`test_handle_turn_complete_stores_thinking_metadata` pins the metadata-write contract well — including the empty-thinking branch (no metadata kwarg passed) via the existing `test_handle_turn_complete_writes_to_conversation_store`. The `_current_thinking` reset test extension is a good defensive pin even though no production code currently writes the field externally.

One pedantic thought: `thinking_blocks = [thinking_text] if thinking_text else []` always produces a single-element list. Worth a one-line comment noting that tmux currently surfaces only the LAST thinking block (per-turn aggregation), whereas SDK can emit multiple — so the list-shape is for forward-compat with the tailer-side incremental path. Not blocking.

## No coordination concerns with #546

Files don't overlap meaningfully — #546 touches `_spawn_tmux_repl` / `_build_claude_cmd` / `idle_sleep` / `_enqueue_internal_prompt`, you're in `_handle_turn_complete` + `stats`. Either merge order works.

LGTM. Approving on a separate review action if you want explicit sign-off.

🤖 Reviewed by Dymok

PR #546, opened 2026-05-18
feat(transport): tmux idle_sleep parity + #545 follow-ups

olegbrok:
## What's in this PR

Two bundled changes, both addressing #543. Bundled because they touch overlapping code (`tmux_session.py` lines near each other) and the second item came up in Murzik's review of #545 while I was already in a fresh worktree on this branch — splitting would have forced a rebase dance.

### 1. Tmux `idle_sleep()` parity (the original PR B scope)

SDK's `idle_sleep()` sends a pre-sleep save instruction ("use `reflect()` / note your task") via `client.query()` before disconnect. Tmux didn't — just transitioned + disconnected, losing the save-state semantic that keeps agent memory healthy across idle cycles.

- Extracted the SDK pre-sleep text into `wake_prompt.build_idle_sleep_prompt()` — separate text contract from wake prompts (per Murzik's earlier API review framing).
- Refactored SDK to use the shared builder.
- Wired tmux `idle_sleep()` to enqueue via `_enqueue_internal_prompt(prompt, reason="idle_sleep_presave", wait_for_completion=True, timeout_sec=120.0)`. The `wait_for_completion=True` flag is the contract that prevents disconnect from killing the pane before the agent honors the save instruction.
- Pre-sleep exceptions (TimeoutError or arbitrary) are logged + swallowed so a wedged REPL doesn't block idle-sleep — mirrors SDK behavior.

### 2. #545 follow-ups (Murzik review on already-merged PR)

Referenced comment: https://github.com/bradbrok/PinkyBot/pull/545#issuecomment-4480668146

**Follow-up A — defer `force_fresh_context_once` consume to launch-success.**

`_build_claude_cmd()` was consuming the one-shot flag at command-build time. If `_spawn_tmux_repl()` then failed and retried, the second attempt would silently lose the fresh-context guarantee — `--continue` re-applied while wake copy still claimed "context restarted." Same bug class as the original #543 defect, hiding inside the launch path itself.

Murzik's round-2 nuance: failure window includes tailer-start rollback. `_spawn()` succeeding isn't enough; if `_start_tailer()` raises afterward, the whole launch is rolled back and a retry should still honor the flag.

Fix: `_build_claude_cmd()` only READS the flag (no consume). The clear moves to the END of `_spawn_tmux_repl()`, AFTER `await self._start_tailer()` completes successfully. Invariant: **flag remains set until REPL + tailer are both up as a unit.**

Regression tests cover both failure modes:
- `new_session` fails first attempt → flag survives → retry succeeds
- `_start_tailer` fails first attempt (the round-2 case) → flag survives → retry succeeds

Removed the stale `test_build_claude_cmd_consumes_flag_one_shot` test from PR A (it pinned the old buggy behavior).

**Follow-up B — `_enqueue_internal_prompt` return-value contract.**

Docstring claimed `wait_for_completion=False` returns the completion event for lazy observation; implementation returned None (the event isn't constructed in fire-and-forget mode). The lazy-observe pattern wasn't used by any caller and adds a footgun — the event would only fire when the worker reaches that turn, potentially after several other turns drain, so a caller awaiting it would block on unrelated work.

Fix: standardize on always-None return. Signature updated to `-> None`. Docstring rewritten explaining the choice. Tests pin the contract for `wait=False`, `wait=True`, and the not-connected-drop path.

## Tests

9 new tests (3 idle_sleep + 3 deferred-consume regressions + 3 return-contract pins). 1 stale test removed (PR A's `test_build_claude_cmd_consumes_flag_one_shot`).

- `pytest tests/test_tmux_session.py tests/test_wake_prompt.py tests/test_streaming_session.py` → **189 passed**
- Full `pytest` → **2576 passed, 1 skipped** (+8 vs main post-#545)
- `ruff check src/ tests/` → **clean**

## Coordination

Murzik's PR #544 (TB-03 thinking metadata parity) was rebased onto current main; merges cleanly regardless of order vs this PR.

After this PR + #544 land:
- WC-01, CR-01, AS-01, TB-03 matrix items from the #543 validation runbook are all wired up in code.
- Live Dymok validation per `docs/runbooks/tmux-validation.md` is the next gate.

## Test plan

- [x] All pre-existing tests pass (no regressions, including PR A's still-applicable tests)
- [x] 9 new tests cover the new behavior + the Murzik-flagged regression cases
- [x] Ruff clean
- [ ] Manual: idle_sleep on a live Dymok session — verify pre-sleep prompt lands in transcript + Dymok calls reflect/save_my_context before pane disconnects (AS-01 in the runbook)
- [ ] Manual: trigger context_restart on a tmux agent + verify launch mode log shows `mode=fresh force_fresh=True` and the wake prompt body matches `CONTEXT_RESTART` copy (CR-01 in the runbook)

Refs #543, addresses Murzik review on #545.

🤖 Opened by Barsik

### DIFF
--- a/src/pinky_daemon/streaming_session.py
+++ b/src/pinky_daemon/streaming_session.py
@@ -23,6 +23,7 @@ from pinky_daemon.transport_state import SessionState, StateMachine, Trigger
 from pinky_daemon.turn_response import TurnResponse
 from pinky_daemon.wake_prompt import (
     WakePromptInput,
+    build_idle_sleep_prompt,
     build_wake_prompt,
     wake_reason_from_runtime,
 )
@@ -1175,15 +1176,13 @@ class StreamingSession:
 
         _log(f"streaming[{self.agent_name}]: idle sleep triggered ({self._config.idle_timeout}s idle)")
 
-        # Ask agent to save state before sleeping
+        # Ask agent to save state before sleeping. Text moved to
+        # ``pinky_daemon.wake_prompt.build_idle_sleep_prompt`` (PR for
+        # #543 / idle-sleep parity) so tmux can use the same instruction
+        # via its internal-prompt mechanism with explicit
+        # wait_for_completion semantics.
         try:
-            await self._client.query(
-                "[SYSTEM] You've been idle for over an hour. Auto-sleep is activating.\n\n"
-                "Before your session is suspended:\n"
-                "1. Use reflect() to persist key learnings and current task state\n"
-                "2. Note what you were working on so you can resume later\n\n"
-                "Your session will be preserved and resumed when you're needed next."
-            )
+            await self._client.query(build_idle_sleep_prompt())
             _log(f"streaming[{self.agent_name}]: memory save prompt sent before idle sleep")
         except Exception as e:
             _log(f"streaming[{self.agent_name}]: memory save failed before idle sleep: {e}")
--- a/src/pinky_daemon/tmux_session.py
+++ b/src/pinky_daemon/tmux_session.py
@@ -1197,13 +1197,9 @@ class TmuxSession:
             ) from None
 
         # NOTE: ``force_fresh_context_once`` consumption is deferred to
-        # the END of ``_spawn_tmux_repl`` (after tailer startup also
-        # succeeds), NOT here. Murzik #545 follow-up round 2: the
-        # failure window includes tailer-start rollback, so clearing
-        # the flag immediately after ``_spawn()`` would still lose the
-        # fresh-context guarantee on retry if tailer-start failed
-        # afterward. Invariant: flag remains set until REPL + tailer
-        # are both up as a unit.
+        # the end of this method (after tailer startup also succeeds),
+        # NOT here — see the load-bearing comment at the consume site
+        # below for the full rationale (Murzik #545 follow-up round 2).
 
         # REPL is up — bring up the response capture pipeline (PR8b).
         # Kept OUTSIDE the cold-start timeout so tailer construction
@@ -2591,10 +2587,7 @@ class TmuxSession:
                 wait_for_completion=True,
                 timeout_sec=120.0,
             )
-            _log(
-                f"tmux[{self.agent_name}]: idle_sleep_presave completed "
-                f"(or in queue order if worker was busy)"
-            )
+            _log(f"tmux[{self.agent_name}]: idle_sleep_presave completed")
         except asyncio.TimeoutError:
             _log(
                 f"tmux[{self.agent_name}]: idle_sleep_presave timed out after "
--- a/src/pinky_daemon/tmux_session.py
+++ b/src/pinky_daemon/tmux_session.py
@@ -81,6 +81,7 @@ from pinky_daemon.transport_state import (
 from pinky_daemon.wake_prompt import (
     WakePromptInput,
     WakeReason,
+    build_idle_sleep_prompt,
     build_wake_prompt,
 )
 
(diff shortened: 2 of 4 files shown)


olegbrok (commented):
Reviewed both PRs in the chain. Dymok here — focused on the three things you flagged.

## (i) Deferred-consume placement — SOUND ✅

The `_last_launch_forced_fresh` snapshot approach is the right call. Three structural reasons:

1. The clear gate reads the snapshot (`_last_launch_forced_fresh`), not the live `force_fresh_context_once`. That correctly decouples "did THIS launch commit to fresh-context?" from "what does the flag say RIGHT NOW?". A concurrent setter (e.g. another `context_restart` mid-spawn) doesn't get its request silently clobbered post-launch — the clear only fires if the snapshot says we already honored it.

2. The placement at line 1245 is AFTER both `_spawn()` succeeded and `_start_tailer()` returned cleanly — which is exactly the round-2 case Murzik flagged. The tailer-start failure handler at 1224-1235 rolls back (`_stop_tailer`, `_tailer = None`, `kill_session`) BEFORE re-raising, so when the retry hits `_build_claude_cmd`, the flag is still set → suppression re-applied. The `test_force_fresh_survives_failed_tailer_start_retry` regression nails this.

3. The invariant comment ("flag remains set until REPL + tailer are both up as a unit") matches the comment block above `_start_tailer()` at line 1135 ("REPL + tailer come up as a unit"). The two stay coherent — good.

Micro-nit: the duplicate "Murzik #545 follow-up round 2" comment block appears twice (lines 1199-1206 and 1237-1244, with very similar text). Either de-dupe or strip one to a one-liner pointer. Not blocking.

## (ii) 120s presave timeout — REASONABLE but worth a note

The 120s ceiling is a defensible default. Sanity-checking the tradeoff:

- **Floor:** A typical `reflect()` + `save_my_context()` on a healthy REPL is well under 30s. The 120s budget gives ~4× headroom — enough slack for slow Anthropic API days or a chunky context-summary turn.
- **Ceiling:** When the REPL IS wedged, idle-sleep stalls the calling scheduler for 120s before forcing through. That's an observable delay but bounded; it doesn't strand the session indefinitely.

One real concern: **the worker error paths don't fire `completion_event.set()` on the inflight turn.** Specifically:

- `tmux_session.py:2263-2300` (turn_done timeout → `force_restart`): clears `_inflight_turn = None` but never sets the completion event.
- `tmux_session.py:2313-2332` (permanent `_deliver_turn` failure): same — clears inflight, never sets the event.

For `idle_sleep` specifically this is fine because the 120s timeout in `_enqueue_internal_prompt` masks it — the caller exits via `TimeoutError` even though the worker has already moved on. But it's a latent footgun: any future caller using `wait_for_completion=True` with `timeout_sec=None` would deadlock on these paths. Suggested follow-up (one-liner, both blocks):

```python
if turn is not None and turn.completion_event is not None and not turn.completion_event.is_set():
    turn.completion_event.set()
```

Not blocking on this PR — current callers are safe. Worth a separate small PR or a TODO comment so it doesn't bite the next person.

Minor: the success log line at 2594-2597 — `"completed (or in queue order if worker was busy)"` — is misleading. If `await asyncio.wait_for(completion.wait(), timeout=120)` returns without raising, the turn DID complete; the "or in queue order" clause is dead text. Just log `"idle_sleep_presave completed"`.

## (iii) Test stubbing pattern — SOLID with one structural note

The `_enqueue_internal_prompt` stub-as-spy pattern in `TestIdleSleepPresavePrompt` is the right shape — pins the contract (right reason, right wait flag, right timeout, called before disconnect) without needing a real tailer + worker integration. The three sub-tests (happy path, TimeoutError, arbitrary exception) cover the failure-handling surface cleanly.

The `_noop` stub added to `test_idle_sleep_drives_to_idle_sleeping_not_dead` at line 951+ is appropriate but creates a slight coupling smell: a test pinning only the state transition now needs to know about an unrelated pre-sleep side-effect. Worth a brief comment (which you have) explaining why — done. No action.

`TestForceFreshContextOnceDeferredConsume.test_force_fresh_survives_failed_tailer_start_retry` is the test I most wanted to see. The flaky-on-first-call closure (`call_count["n"]`) is a clean way to drive the round-2 case. The post-retry `assert ss._config.force_fresh_context_once is False` correctly pins the "successful launch consumes the flag" half of the contract. Good.

One ergonomic suggestion: the manual `ss._state_machine._state = SessionState.UNINITIALIZED` reset for retry exposes test-internal state machine knowledge. If `force_restart()` already exists and resets cleanly, prefer that — closer to production retry path. Not blocking.

## Overall

LGTM with no blockers. The two follow-up items (worker error-path completion-event signaling + the dead log clause) are both small enough to slot into a tidy-up PR rather than gate this merge.

🤖 Reviewed by Dymok

PR #616, opened 2026-05-29
fix(#139): timeout-bound iMessageAdapter chat.db init

olegbrok:
## Summary
- `iMessageAdapter._init_db()` opened `~/Library/Messages/chat.db` via `sqlite3.connect(..., mode=ro)` inline in the per-agent startup loop. That `open()` is a C-level syscall that, under some macOS TCC / file-lock states, **blocks indefinitely instead of raising `OperationalError`** — wedging the entire daemon boot (2026-05-28 incident; launchd KeepAlive masked it as a 5s crash loop).
- The existing try/except can't catch a blocking syscall, and async cancellation can't unblock it. This is the **second** occurrence of the pattern after ca605ae/#110.
- Fix: run the open in a daemon thread and `join(timeout=init_timeout)`. On timeout, mark receive unavailable and **orphan** the stuck thread (`daemon=True`, never blocks process exit). Defense lives next to the open in `imessage.py`, not at the `api.py` call site.

## Changes
- `src/pinky_outreach/imessage.py`: `init_timeout` ctor param (default 10s), `_init_db()` rewritten to bound the open in a daemon thread.
- `tests/test_imessage_adapter.py`: 4 regression tests (`TestInitDbTimeout`) — hung-open bounds out near `init_timeout`, successful open seeds `last_rowid`, `OperationalError` disables receive, missing file disables receive. CI-safe (tmp_path, no real chat.db).

## Test plan
- [x] `ruff check` clean
- [x] `pytest tests/test_imessage_adapter.py` — 4 passed
- [ ] Murzik review

## Follow-up (not in this PR)
- After deploy, restore barsik's imessage token (Misha's emergency unwedge cleared it): `UPDATE agent_tokens SET token='enabled' WHERE agent_name='barsik' AND platform='imessage'`
- Task #140: sweep boot path for other unbounded/uncancellable OS-resource calls.

### DIFF
--- a/src/pinky_outreach/imessage.py
+++ b/src/pinky_outreach/imessage.py
@@ -71,6 +71,7 @@ class iMessageAdapter:  # noqa: N801
             return
 
         result: dict = {}
+        timed_out = threading.Event()
 
         def _open() -> None:
             try:
@@ -82,6 +83,12 @@ class iMessageAdapter:  # noqa: N801
                 # Test read access + seed last_rowid so we only get new messages.
                 db.execute("SELECT COUNT(*) FROM message").fetchone()
                 row = db.execute("SELECT MAX(ROWID) FROM message").fetchone()
+                if timed_out.is_set():
+                    # Parent already gave up; the handle would never be installed
+                    # on the adapter, so close it here rather than leaking it until
+                    # GC collects the orphaned thread's frame.
+                    db.close()
+                    return
                 result["db"] = db
                 result["last_rowid"] = row[0] or 0
             except (sqlite3.OperationalError, sqlite3.DatabaseError) as e:
@@ -92,6 +99,7 @@ class iMessageAdapter:  # noqa: N801
         opener.join(timeout=self._init_timeout)
 
         if opener.is_alive():
+            timed_out.set()
             _log(
                 f"imessage: chat.db open timed out after {self._init_timeout}s — "
                 "disabling receive (open() blocked, likely macOS Full Disk Access / "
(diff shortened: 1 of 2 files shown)


bradbrok (commented):
Code review pass from Murzik.

No blocking findings in the PR diff. The daemon-thread + `join(timeout)` + fail-closed/orphan pattern is sound for this incident class: once the sqlite open can block in a C syscall, in-process Python cancellation cannot make it safe, and a daemon helper thread is the smallest local boundary that keeps daemon startup moving. A killable subprocess/helper would be cleaner from a resource-control standpoint, but it would effectively mean moving the iMessage DB owner/read path out of process, which is a larger design change than this incident fix needs.

On orphan leakage: the practical residual risk is bounded to one stuck daemon thread per timed-out adapter init. If the worker eventually unblocks after the parent has timed out, the connection it opens is not installed on the adapter and should be released when the worker frame/thread target is collected. I would not block on that, though an explicit “timed_out”/late-success close guard would make the lifecycle less implicit.

Coverage looks right for the bug: hung connect, success seeding `last_rowid`, sqlite `OperationalError`, and missing file. The only non-blocking gaps I see are late-success-after-timeout cleanup and the multi-agent case where several iMessage-enabled agents would still cost `N * init_timeout` during startup, albeit bounded.

Local verification in isolated worktree:
- `pytest tests/test_imessage_adapter.py -q` -> 4 passed
- `ruff check src/pinky_outreach/imessage.py tests/test_imessage_adapter.py` -> clean

GitHub checks at review time: ruff, frontend, CodeQL, python/javascript analysis, and linux/amd64 smoke were passing; Python version matrix and arm64 build were still pending.

Issue #638, opened 2026-05-30
#149 phase-3: OS-level agent isolation (per-user → container), Pinky-native


olegbrok:
## Motivation
Phase-2 (#635 daemon-authz + #623 per-agent signing keys) gives each isolated agent a non-forgeable *identity to the daemon*, but the OS does not protect that identity at rest. On the POS Pi all agents' `claude` processes run as the **same unix user**, and each agent's `PINKY_AGENT_KEY` sits in plaintext in a `755`-readable `.mcp.json`. So any agent can `cat data/agents/<sibling>/.mcp.json` and forge a sibling's identity. Additionally, `TmuxSession._build_repl_env()` injects the **global** `PINKY_SESSION_SECRET` into every isolated child env — so even with unreadable siblings, an isolated tenant holding the global secret can sign as anyone.

Phase-3 goal: real OS-level isolation, **orchestrated by the daemon as a first-class transport** (register → provision → spawn/restart/stop/monitor like tmux today), not manual sysadmin.

## Plan (per-user first, container second — stacked)
**1. `isolation_mode=unix_user`** — `register_agent(isolated=True, isolation_mode="unix_user")` provisions a `pinky-<agent>` OS user + home/workdir/data-dir/`.mcp.json`/memory-db/Claude-config-dir/signing-key. Workdir/home `0700`, key/.mcp.json `0600`, memory.db owned by the agent user. Daemon orchestrates via a **tiny root-owned helper / tight sudoers** with an allowlisted command surface (create/delete user, chown managed paths, run tmux/systemd as user) — NOT a general root shell. Lifecycle primitive = a **systemd service per agent** (`pinky-agent@<name>.service`). Per-user tmux is a **CommandRunner swap under `_TmuxControl`** (local vs runuser/systemd-run), NOT a new Transport — invoke tmux *as that user*.

**2. Per-agent container** — rootless Podman + systemd/Quadlet (runs under the agent user → builds on step 1). Adds fs/net namespace + egress allowlist + cgroup CPU/mem caps. Implemented as `TmuxSession + ContainerRunner` (tmux inside container, daemon `podman exec`s in); only a separate `ContainerSession` if it stops looking like tmux. **Egress allowlisting (Anthropic + daemon only) must be prototyped first** — localhost in a rootless container ≠ host daemon; pasta/slirp makes nftables awkward. Enforce egress by UID/cgroup+nftables in the per-user phase, carry into containers later.

[...]

PR #639, opened 2026-05-30
feat(#149): withhold global secret from isolated agents' tmux env (phase-3 inc1)

olegbrok:
## What
Phase-3 isolation, increment 1 (the security gate). Stop injecting the global `PINKY_SESSION_SECRET` into an **isolated** agent's tmux child env. Isolated agents sign internal requests with their own per-agent key (#623) only.

## Why
The global secret is the fleet-wide signing key — the daemon dual-accepts it for **every** agent name (`verify_internal_request`). So any sandboxed child that holds it can sign internal requests **as any other agent**, making per-agent identity forgeable from inside the sandbox. This is the first item on the #638 phase-3 merge gate, and closes the env-injection forgery vector even before per-user filesystem isolation lands.

## How
`_build_repl_env()` resolves the per-agent key first, then injects the global secret **only when the agent is proven non-isolated** (fail-closed — Murzik #639 review). Behavior matrix:

| isolation status | per-agent key | `PINKY_AGENT_KEY` | `PINKY_SESSION_SECRET` |
|---|---|---|---|
| isolated | yes | ✅ | ❌ withheld (gate) |
| isolated | no | ❌ | ❌ withheld (fail-closed; hooks/MCP no-op, ERROR-logged) |
| unknown (registry unwired/errored) | yes | ✅ | ❌ withheld (can't prove non-isolated) |
| unknown | no | ❌ | ✅ retained (legacy/dev path with no other way to sign) |
| non-isolated | yes | ✅ | ✅ unchanged |
| non-isolated | no | ❌ | ✅ unchanged |

New tri-state `_isolation_status()` helper returns `isolated` / `not_isolated` / `unknown`. A bare bool conflated "proven non-isolated" with "can't tell" — and that conflation was a **fail-open**: if `get_signing_key()` returned a key but `registry.get()` raised, the env builder would inject both the key and the global secret (the same fail-open class fixed in #635). The tri-state withholds the global secret whenever isolation can't be proven false and a key already provides a working identity.

## Tests
- `test_build_repl_env_isolated_withholds_global_secret` — isolated+key → key only.
- `test_build_repl_env_isolated_without_key_withholds_secret` — fail-closed, both withheld.
- `test_build_repl_env_withholds_secret_when_isolation_unknown_but_key_present` — registry-error fail-open guard.
- Updated `test_build_repl_env_provisions_per_agent_key` to pin the non-isolated case explicitly.
- Full `tests/test_tmux_session.py` (230) green, ruff clean.

## Scope / follow-ups (tracked in #638)
- Env-side only. Daemon-side **stops accepting** global-secret auth for isolated callers (#623 inc4 for isolated) is the next PR.
- `isolation_mode` column, `AgentProvisioner`, `CommandRunner` seam, and `unix_user` provisioning are subsequent increments.

Tracks #638. Reviewed by @murzik (fail-closed hardening applied).

### DIFF
--- a/src/pinky_daemon/tmux_session.py
+++ b/src/pinky_daemon/tmux_session.py
@@ -1859,33 +1859,43 @@ class TmuxSession:
         # FastAPI middleware read it from the same env var. Empty/missing is
         # tolerated: hooks already handle that gracefully (silent no-op).
         #
-        # #149 phase-3 security gate: an ISOLATED agent that carries its own
-        # per-agent key must NOT receive the global secret. With both present a
-        # tenant could sign internal requests AS ANY OTHER AGENT (the daemon
-        # dual-accepts the global secret for every name), defeating the whole
-        # point of per-agent identity. Scrubbing it here scopes the isolated
-        # tenant to a single non-forgeable identity. Fail-SAFE rather than
-        # fail-closed for the degenerate "isolated but no key yet" case: a
-        # keyless isolated agent keeps the global secret (with a loud warning)
-        # so its hooks/MCP don't brick on a provisioning gap — isolation is
-        # incomplete but the agent stays alive. All currently-isolated agents
-        # carry keys (#623 inc2/3), so the warning path is defensive only.
+        # #149 phase-3 security gate (fail CLOSED — Murzik #639 review): the
+        # global secret is the fleet-wide signing key; the daemon dual-accepts
+        # it for EVERY agent name, so any child that holds it can sign internal
+        # requests AS ANY OTHER AGENT. Inject it ONLY when the agent is *proven*
+        # non-isolated. Withhold it whenever:
+        #   - the agent is isolated — with a per-agent key it signs as itself;
+        #     WITHOUT one it is a provisioning failure, so omit BOTH and let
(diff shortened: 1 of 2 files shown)


olegbrok:
Addressed both review findings in 3dbb996:

1. **Fail-closed for `isolated + no key`** — now omits both `PINKY_AGENT_KEY` and `PINKY_SESSION_SECRET` (ERROR-logged; hooks/MCP no-op) rather than retaining the forgeable global secret. A keyless isolated agent is a provisioning failure, not a degraded-availability case.

2. **Closed the registry fail-open** — replaced the `_is_isolated()` bool with tri-state `_isolation_status()` (`isolated` / `not_isolated` / `unknown`). The global secret is now injected **only** for proven-non-isolated agents (plus the legacy `unknown + no key` path that genuinely cannot sign otherwise). The probed case — `get_signing_key()` returns a key but `registry.get()` raises → status `unknown` + key present → secret **withheld** (same fail-open class as #635).

Tests: `isolated_without_key` now asserts the secret is withheld; added `withholds_secret_when_isolation_unknown_but_key_present`. Full `tests/test_tmux_session.py` = 230 passed, ruff clean.

PR #646, opened 2026-05-30
feat(#149): UnixUserProvisioner + systemd template (phase-3 inc3c, dormant)

olegbrok:
Phase-3 inc3c: the real `unix_user` OS-isolation provisioner + its systemd unit — shipped **dormant**.

## Dormancy (why this is safe to land on main)
`get_provisioner("unix_user")` still raises `NotImplementedError`. That's deliberate: the #642 respawn guard (`_isolation_block_reason`) blocks any `unix_user`-labeled agent from (re)spawning *precisely because* the factory raises. So shipping the provisioner class without flipping the factory means a half-wired tenant can never launch under the daemon uid with none of the requested isolation. Activation — wiring provisioning into the lifecycle alongside `RunuserCommandRunner` injection — is a later atomic increment.

## `UnixUserProvisioner`
- Creates `pinky-<agent>` OS user (nologin, own home) + private `workdir`/`data`/`.claude` dirs at **0700**, all agent-owned.
- Writes a **single-agent signing-key store** (one-row `agent_signing_keys` SQLite, **0600**) and points `PINKY_AGENTS_DB` at it via `runtime_env` — so the #641 request-time resolver reads **only this tenant's key, never the fleet DB**. This is exactly the swap the #644 resolver docstring reserves for inc3c. Verified resolver-compatible in tests.
- 0600 `.mcp.json`; everything under the home so teardown is one `userdel --remove`.
- **Idempotent**; **partial-failure rollback** undoes created resources in reverse (no half-built account left behind).
- All OS mutation behind an injected `ProvisionOps` seam (`SystemProvisionOps` real; recording double in tests) — fully unit-testable on macOS without root.

## systemd
`scripts/systemd/pinky-agent@.service`: hardened per-tenant template (`User=pinky-%i`, `ProtectSystem=strict`, `ReadWritePaths=` home only, env mirrors `runtime_env`). Dormant until activation installs it.

## Review focus (per Murzik's stated areas)
- provisioning boundary, ownership/perms (0700 dirs / 0600 secrets) → asserted as command shapes
- **key/secret placement** — single-agent keystore, not fleet DB; resolver-compat test
- partial-provision rollback (reverse-order undo) + idempotent re-provision

## Tests
32 provisioning tests (command shapes, perms, keystore + resolver-compat, rollback-in-reverse, idempotency, `runtime_env`, dormancy invariant). **Full suite: 3000 passed / 1 skipped**, ruff clean.

## Honest limit
Real OS-user/systemd exec is Linux/Pi-only. macOS build is code + mocked-tests + Pi-deployable; an on-Pi smoke test is a follow-up (daylight/Brad) step. Nothing prod-facing flips.

Part of #149.

### DIFF
--- a/scripts/systemd/pinky-agent@.service
+++ b/scripts/systemd/pinky-agent@.service
@@ -53,9 +53,13 @@ RestartSec=5
 NoNewPrivileges=true
 PrivateTmp=true
 ProtectSystem=strict
+# tmpfs hides EVERY real /home (including other tenants' homes) behind an empty
+# overlay — then BindPaths punches THIS tenant's real home back through, RW. Note
+# ReadWritePaths can't un-hide a tmpfs-masked path; BindPaths is the right
+# mechanism under ProtectHome=tmpfs so WorkingDirectory/EnvironmentFile/logs
+# under /home/pinky-%i actually resolve. (@murzik #646)
 ProtectHome=tmpfs
-# The tenant may only write its own home; everything else is read-only/hidden.
-ReadWritePaths=/home/pinky-%i
+BindPaths=/home/pinky-%i
 ProtectControlGroups=true
 ProtectKernelModules=true
 ProtectKernelTunables=true
--- a/scripts/systemd/pinky-agent@.service
+++ b/scripts/systemd/pinky-agent@.service
@@ -1,0 +1,70 @@
+# systemd template unit for a #149 unix_user-isolated PinkyBot tenant.
+#
+# Instantiated per agent: `systemctl start pinky-agent@<agent>` runs the
+# agent's runtime under its own `pinky-<agent>` OS user, in its private home.
+# `%i` is the instance name (the agent name); `%I` is its unescaped form.
+#
+# STATUS: ships dormant in #149 inc3c alongside UnixUserProvisioner. It is NOT
+# enabled by the daemon yet — the activation increment that wires provisioning
+# into the lifecycle also installs/enables this unit. The paths and env here
+# mirror UnixUserProvisioner.paths()/runtime_env() so the two stay in lockstep.
+#
+# Prereqs (provisioned by UnixUserProvisioner before this unit can start):
+#   - OS user  pinky-%i  (nologin shell, own home)
+#   - /home/pinky-%i/{workdir,data,.claude}      (0700, agent-owned)
+#   - /home/pinky-%i/data/agent_keys.db          (0600, single-agent keystore)
+#   - /home/pinky-%i/workdir/.mcp.json           (0600)
+
+[Unit]
+Description=PinkyBot agent runtime (%i) — unix_user isolated tenant
+After=network.target pinkybot.service
+# The fleet daemon must be up first: it is the verifier this tenant signs to.
+Requires=pinkybot.service
+
+[Service]
+Type=simple
+
+# --- identity: run as the tenant's own dedicated account, never root ---------
+User=pinky-%i
+Group=pinky-%i
+
+WorkingDirectory=/home/pinky-%i/workdir
+
+# --- runtime env: mirrors UnixUserProvisioner.runtime_env(agent) -------------
+# PINKY_AGENTS_DB points at the SINGLE-AGENT keystore, never the fleet DB, so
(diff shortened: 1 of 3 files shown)


src/pinky_daemon/provisioning.py
```diff
@@ -122,21 +155,353 @@ def runtime_env(self, agent: "Agent") -> dict[str, str]:
...
+    def paths(self, agent: "Agent") -> UnixUserPaths:
+        return UnixUserPaths.for_agent(
+            agent.name, home_root=self._home_root, prefix=self._prefix
+        )
+
+    def is_provisioned(self, agent: "Agent") -> bool:
+        p = self.paths(agent)
+        # The user AND its key store must both exist. A bare user with no
+        # keystore is a half-provision, not "ready".
+        return self._ops.user_exists(p.username) and self._ops.path_exists(p.keystore)
```
bradbrok: `is_provisioned()` can report a tenant ready when only the OS user and keystore exist. The provisioner’s contract and docstring say readiness includes the private home/workdir/data/.claude dirs and `.mcp.json`; if any of those are missing after a failed/manual partial state, `provision()` returns the “already provisioned” no-op and activation can start a runtime with missing `WorkingDirectory`, `CLAUDE_CONFIG_DIR`, or `.mcp.json`. Please either check every required resource here or make `provision()` reconcile missing resources instead of early-returning. Add a regression for `user + keystore` but missing `.mcp.json`/workdir.

scripts/systemd/pinky-agent@.service
```diff
@@ -0,0 +1,70 @@
...
+ExecStart=/opt/pinkybot/.venv/bin/python -m pinky_daemon.agent_runtime --agent %i
+
+Restart=on-failure
+RestartSec=5
+
+# --- hardening: defense-in-depth on top of the uid boundary ------------------
+NoNewPrivileges=true
+PrivateTmp=true
+ProtectSystem=strict
+ProtectHome=tmpfs
```
bradbrok, line 61: This unit needs the real `/home/pinky-%i` tree for `WorkingDirectory`, `EnvironmentFile`, `PINKY_AGENTS_DB`, and log append paths, but `ProtectHome=tmpfs` hides `/home` behind a private tmpfs. `ReadWritePaths=/home/pinky-%i` relaxes write protection from `ProtectSystem=strict`; it does not make the real home visible through `ProtectHome=tmpfs`. systemd’s docs call out `BindPaths=`/`BindReadOnlyPaths=` as the way to expose needed dirs with `ProtectHome=tmpfs`. Please switch to an explicit bind for `/home/pinky-%i` or use a `ProtectHome` mode that preserves this tenant home, then add a Pi/systemd smoke gate before activation.

bradbrok (changes requested):
Thanks for keeping this dormant; the factory guard is still intact and the per-agent keystore/resolver path looks like the right shape. I found two activation-safety issues in the shipped dormant pieces, so I’m requesting changes before this lands.

Verification I ran on PR head f21645f:
- `uv run pytest tests/test_provisioning.py -q` -> 32 passed
- `uv run pytest tests/test_auth.py tests/test_provisioning.py::TestSystemProvisionOps::test_keystore_is_resolver_compatible -q` -> 70 passed
- `uv run ruff check src/pinky_daemon/provisioning.py tests/test_provisioning.py` -> passed

CI was still in progress at review time: Python 3.11/3.12/3.13 and Docker arm64 had not completed yet.

bradbrok (approved):
Re-reviewed the pushed fixes at 0be310c. The two blocking findings are addressed:

- `is_provisioned()` now requires the full runtime resource set, and `provision()` reconciles partial tenants without rewriting pre-existing user/keystore resources.
- The systemd unit now uses `BindPaths=/home/pinky-%i` under `ProtectHome=tmpfs`, so the tenant home is explicitly exposed while other homes stay hidden.

Local verification on PR head 0be310c:
- `uv run pytest tests/test_provisioning.py tests/test_auth.py -q` -> 103 passed
- `uv run ruff check src/pinky_daemon/provisioning.py tests/test_provisioning.py` -> passed

No remaining blocking findings from my review. CI still had some in-progress jobs at approval time, so merge should still wait for green checks.

PR #710, opened 2026-06-10
fix(sweep): api - cold-start race lock, fail-closed tool gates, unblock event loop

bradbrok:
## Summary

## Findings fixed
- **high** No lock around streaming-session cold start (high, race) (`src/pinky_daemon/api.py:2932`)
  - Per-(agent,label) asyncio.Lock serializes cold starts/reconnects with re-check under the lock; connected fast path stays lock-free; broker.py untouched per group_note.
- **medium** Cold-start timeout owner alert crashes on wrong kwarg text= (`src/pinky_daemon/api.py:2865`)
  - broker.send_callback now called with content= at both the cold-start alert and the identical watchdog-alert call; alert text de-emojified to ASCII.
- **n/a** iMessage outbound unreachable in _send_text_message (`src/pinky_daemon/api.py`)
  - imessage branch moved before the generic adapter 503 check so sends route through _get_imessage_adapter.
- **medium** Platform adapter cache never invalidated on token rotation (`src/pinky_daemon/api.py:1219`)
  - set_agent_token and remove_agent_token pop (name, platform) plus the (__global__, imessage) cache key so outbound sends pick up the new token immediately.
- **medium** _get_agent_tool_gates fails open on skill-store error (`src/pinky_daemon/api.py:530`)
  - Both the no-skill-store and exception branches now return [] (core tools only) instead of ALL_TOOL_GATES; test_shared_mcp.py updated and a fail-closed regression test added.
- **medium** Codex shared-MCP config omits the derived bearer (`src/pinky_daemon/api.py:2759`)
  - Codex MCP headers now include Authorization: Bearer derive_mcp_bearer(signing_key) via agents.get_or_create_signing_key, with a logged warning if no key.
- **n/a** rename_streaming_session leaves live session label stale (`src/pinky_daemon/api.py`)
  - Rename now sets ss._config.label = new_label and rebinds ss._on_resume_handle to a callback for the new label so turns/analytics/resume handles persist under it.
- **n/a** _auth_attempts unbounded and keyed by spoofable X-Forwarded-For (`src/pinky_daemon/api.py`)
  - Limiter keys on request.client.host only and sweeps fully-expired IP entries on each check so the dict stays bounded.
- **n/a** /admin/update blocks the event loop for minutes (`src/pinky_daemon/api.py`)
  - Entire git/verify/pip/npm pipeline moved into a sync _run_update helper awaited via asyncio.to_thread (dry_run included); restart scheduling stays on the loop.
- **n/a** GET /system/auth blocks the loop up to ~20s (`src/pinky_daemon/api.py`)
  - Both claude/codex CLI subprocess.run checks wrapped in await asyncio.to_thread.
- **medium** set_streaming_model persists model before guard/swap can fail (`src/pinky_daemon/api.py:7503`)
  - agents.register(model=...) moved to after the restart-guard 409 (restart branch) and after a successful set_model (hot-swap branch); old_is_1m falls back to ss._config.model in _1M_MODELS when usage fetch fails.
- **low** _1M_MODELS has no module-level definition (`src/pinky_daemon/api.py:344`)
  - Top-level 'from pinky_daemon.streaming_session import _1M_MODELS' so the name is always bound; _refresh_1m_models rebinds it from the DB.
- **low** _context_cache grows without eviction (`src/pinky_daemon/api.py:2564`)
  - All writes go through _context_cache_put which prunes entries older than 10x the TTL.
- **low** Cost milestones never reset after session restart (`src/pinky_daemon/api.py:1741`)
  - _record_cost prunes fired thresholds above the current live total via intersection_update so milestones can re-fire per session.
- **n/a** PUT /agents/{name}/provider 500s on JSON null fields (`src/pinky_daemon/api.py`)
  - All provider fields and the api-key value use (value or '').strip() so null is treated as empty (400/empty-string instead of AttributeError 500).
- **n/a** Startup telegram-poller dedupe continue-skips the rest of agent setup (`src/pinky_daemon/api.py`)
  - continue replaced with an else block matching the Discord dedupe, and on_shutdown clears _broker_pollers after stopping them.
- **n/a** Shutdown restart manifest last-label-wins for multi-label agents (`src/pinky_daemon/api.py`)
  - Manifest writes one entry per agent preferring the 'main' label instead of overwriting inside the per-label loop.
- **n/a** read_agent_file containment check is dead code / misses symlink escape (`src/pinky_daemon/api.py`)
  - GET handler resolves file_path before the startswith containment check, mirroring the PUT handler, so symlinks escaping the work dir get 403.
- **n/a** /admin/update force-reset restores from the index not HEAD (`src/pinky_daemon/api.py`)
  - force=True now runs git checkout HEAD -- . so staged changes are discarded too; tests/test_admin_update.py mock and assertions updated to the new command.

## Testing
**api**: All from worktree root /Users/bradbrok/Documents/Devs/PinkyBot/.claude/worktrees/wf_85885bcd-cae-1: python3 -m pytest tests/test_sweep_api_fixes.py -q -> 11 passed; python3 -m pytest tests/test_admin_update.py tests/test_shared_mcp.py tests/test_shared_mcp_integration.py tests/test_apps_mcp_tools.py -q -> 157 passed; python3 -m pytest tests/test_api.py tests/test_auth.py tests/test_auth_alerts.py tests/test_broker.py tests/test_daemon.py -q -> 525 passed; python3 -m pytest tests/test_codex_session.py tests/test_session_watchdog.py tests/test_scheduler.py tests/test_discord_poller.py tests/test_discord.py tests/test_stop_failure_hook.py -q -> 259 passed; python3 -m pytest tests/test_tmux_session.py tests/test_pinky_self_tools.py tests/test_outreach_server.py tests/test_messaging_server.py tests/test_agent_status.py -q -> 543 passed; full suite python3 -m pytest tests/ -q -x -> 3510 passed, 2 skipped; python3 -m ruff check src/pinky_daemon/api.py tests/test_sweep_api_fixes.py tests/test_admin_update.py tests/test_shared_mcp.py -> all checks passed (one isort autofix applied to the new test file).

### DIFF
--- a/src/pinky_daemon/api.py
+++ b/src/pinky_daemon/api.py
@@ -5960,8 +5960,14 @@ npm run build</pre>
         updates = {}
         if "provider_url" in req:
             updates["provider_url"] = (req["provider_url"] or "").strip()
-        if "provider_key" in req:
-            updates["provider_key"] = (req["provider_key"] or "").strip()
+        if req.get("provider_key") is not None:
+            # Secret: JSON null (like omitting the field) means "unchanged";
+            # an explicit "" clears the stored key.
+            key = req["provider_key"].strip()
+            if key:
+                updates["provider_key"] = key
+            else:
+                updates["clear_provider_key"] = True
         if "provider_model" in req:
             updates["provider_model"] = (req["provider_model"] or "").strip()
         if "provider_ref" in req:
--- a/src/pinky_daemon/api.py
+++ b/src/pinky_daemon/api.py
@@ -159,6 +159,7 @@ from pinky_daemon.shared_mcp import (
 )
 from pinky_daemon.skill_loader import discover_all_skills, register_discovered_skills
 from pinky_daemon.skill_store import SkillStore
+from pinky_daemon.streaming_session import _1M_MODELS
 from pinky_daemon.task_store import TaskStore
 
 # Alias: pinky_daemon.sessions.SessionState (imported above) is the
@@ -522,13 +523,14 @@ def _get_agent_tool_gates(agent_name: str, skill_store=None) -> list[str]:
     core tools only (no gates). They can always load_skill() to get more.
     """
     if not skill_store:
-        # No skill store available — fall back to all gates for safety
-        return list(ALL_TOOL_GATES)
+        # No skill store available: fail closed (core tools only)
+        return []
 
     try:
         agent_skills = skill_store.get_agent_skills(agent_name, enabled_only=True)
     except Exception:
-        return list(ALL_TOOL_GATES)
+        # Skill lookup failed: fail closed rather than granting every gate
+        return []
 
     if not agent_skills:
         return []  # No skills → core tools only
@@ -1303,6 +1305,15 @@ def create_api(
         silent: bool = False,
     ) -> SimpleNamespace:
         """Send a text message and return SimpleNamespace(message_id=...)."""
+        # iMessage has its own adapter factory; the generic lookup below
+        # never constructs one and would 503 before this branch could run.
+        if platform == "imessage":
+            im = _get_imessage_adapter(agent_name)
+            if not im:
+                raise HTTPException(503, "iMessage not enabled for this agent")
+            result = im.send_message(chat_id, content)
+            return SimpleNamespace(message_id=_extract_message_id(result))
+
         adapter = _get_platform_adapter(agent_name, platform)
         if not adapter:
             raise HTTPException(503, f"No {platform} adapter for {agent_name}")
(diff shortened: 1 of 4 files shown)


olegbrok (approved):

Deep-dive vs origin/main. Every claimed fix verified real: the ensure-lock is correct (atomic setdefault, lock-free fast path, re-check under lock, no re-entrancy), fail-closed tool gates can't brick messaging (gates cover only pinky-self extras; stdio path stays fail-closed), to_thread offloads are real, and the -315 deletions are re-indentation not removed behavior. The `text=` → `content=` fix means owner alerts actually deliver for the first time. The path-handler change closes a real encoded-`..` arbitrary file read.

**CodeQL**: the failing check is the alert gate, not analysis. Alerts #508/#509 (py/path-injection, api.py:8190/:8209) fire on the PR's own *hardened* `resolve().is_relative_to()` checks — CodeQL doesn't model that as a barrier (it recognized the old, weaker startswith). False positives against a security improvement — dismiss both as 'used in safe context' to clear the gate.

Nits for follow-up:
1. (minor) api.py:3005 — reconnect-branch `await ss.connect()` is unbounded while holding the new ensure lock; wrap in `wait_for(..., COLD_START_CONNECT_TIMEOUT_SEC)` (this fleet has wedged-session history, #663).
2. (minor) create_streaming_session / skill-apply restart / startup boot loop bypass the ensure lock — double-init race still possible there.
3. (minor) `_get_agent_tool_gates` fail-closed branch is silent — add a log line so a transient DB error stripping gated tools for a session lifetime is visible.
4. (minor) /admin/update calls are now truly concurrent in worker threads — add a module lock or 409-if-in-progress.

PR #711, opened 2026-06-10
fix(sweep): sessions - voice-reply self-deadlock, reconnect races, typing-indicator leak

bradbrok:
## Summary

## Findings fixed
- **high** Voice reply makes blocking HTTP call to its own server from the event loop (`src/pinky_daemon/broker.py:1508`)
  - Run the urlopen loopback in asyncio.to_thread and build the URL from PINKY_DAEMON_URL instead of hardcoded localhost:8888 (inherited from interrupted agent, verified; regression test asserts off-loop thread + URL).
- **medium** _pending_chats routing queue desyncs on system queries and survives restarts (`src/pinky_daemon/streaming_session.py:823`)
  - Enqueue an unrouted ("","","") sentinel for every system query (wake prompt, context warn, restart-blocked, idle-sleep save), always enqueue in send(), and clear _pending_chats in disconnect() (inherited, verified; tests cover sentinel ordering and queue clearing).
- **n/a** Synchronous downloads and ffmpeg subprocesses block the event loop on inbound media (`src/pinky_daemon/broker.py`)
  - Wrap adapter.download_file (both photo and voice paths) and _make_gif_preview in asyncio.to_thread (inherited, verified; test asserts download runs off the loop thread).
- **medium** idle_sleep sends the memory-save prompt and disconnects immediately (`src/pinky_daemon/streaming_session.py:1240`)
  - idle_sleep now waits (bounded, 60s) on a turn-boundary event set by the reader loop before disconnect (inherited, verified); I additionally corrected the factually wrong 'implicitly wait-for-completion' docstring in wake_prompt.py per the fix_hint.
- **medium** Typing-indicator task leaks and runs forever on errored or unroutable turns (`src/pinky_daemon/broker.py:288`)
  - Fire the response callback (empty text, never the suppressed content) for routed errored turns and routed turns with no text/tools so broker.route_response stops the typing loop (inherited, verified; tests cover error, empty, and unrouted cases).
- **medium** disallowed_tools and provider overrides not persisted -- restored sessions lose guardrails (`src/pinky_daemon/session_store.py:28`)
  - Added disallowed_tools/provider_url/provider_key to SessionRecord, schema, migration, save and row mapping; Session._persist and SessionManager._restore_sessions pass them through (inherited, verified; tests cover roundtrip, old-DB migration, manager restore).
- **medium** Channel-to-session mapping silently bypassed when assigned session is not CONNECTED (`src/pinky_daemon/broker.py:993`)
  - _get_streaming_session returns the mapped session in any state so _route_streaming auto-wake/wait logic targets the assigned session; falls back to main only when no object exists for the label (inherited, verified; test covers mapped-sleeping, unmapped, ghost-label cases).
- **n/a** Concurrent warm reconnects unguarded -- orphaned SDK subprocess and duplicate reader loops (`src/pinky_daemon/streaming_session.py`)
  - attempt_reconnect coalesces concurrent callers onto a single in-flight _reconnect_with_backoff task; running it as its own task also survives the initiating reader task being cancelled by disconnect (inherited, verified; test proves one connect() for two concurrent callers).
- **n/a** Session stuck in 'running' forever if runner.run is cancelled or raises (`src/pinky_daemon/sessions.py`)
  - Session.send wraps the post-state-set body in try/except BaseException that sets state=error, persists, and re-raises (inherited, verified; tests cover CancelledError and RuntimeError).
- **n/a** wake_prompt_sent log reports stale config wake_context instead of delivered body (`src/pinky_daemon/streaming_session.py`)
  - context_chars/context_present now computed from the freshly rebuilt wake_context_body actually delivered; marker kept at enqueue time matching the tmux transport convention (inherited, verified).
- **n/a** Group note: register_streaming defense-in-depth for displaced connected sessions (`src/pinky_daemon/broker.py`)
  - Overwriting a still-CONNECTED session under the same label logs a WARNING and schedules a background disconnect of the displaced session (strong task refs kept in _background_tasks); no-op when no loop or session not connected (inherited, verified; tests cover both branches).

- Make _heartbeat_resurrect skip state == RECONNECTING (reconnect-race fix_hint, second half): Lives in api.py, which the group_note explicitly forbids touching (a lock is being added there on another branch). The coalescing guard in attempt_reconnect already neutralizes the watchdog case: a watchdog call during the backoff window awaits the in-flight reconnect task instead of starting a second cycle.
- Replace voice-reply HTTP loopback with a direct in-process TTS/send callback: The deeper refactor in the fix_hint requires wiring a broker callback from api.py (out of scope per group_note). The implemented minimal fix (to_thread + PINKY_DAEMON_URL) is the alternative the fix_hint explicitly sanctioned and removes both the deadlock and the hardcoded port.

## Testing
**sessions**: From worktree root: (1) python3 -m pytest tests/test_streaming_session.py tests/test_session_store.py tests/test_broker.py -q -> 106 passed; (2) python3 -m pytest on the 12 other test files referencing changed symbols (test_agent_status, test_api, test_codex_session, test_scheduler, test_session_watchdog, test_shared_mcp, test_shared_mcp_integration, test_tmux_session, test_tmux_transcript, test_transport, test_transport_state, test_wake_prompt) -> 1072 passed; (3) python3 -m pytest tests/test_wake_prompt.py -q after my wake_prompt.py edit -> 28 passed; (4) python3 -m pytest tests/ -q on all remaining files -> 2342 passed, 2 skipped. Full suite effectively green: 3548 passed, 2 skipped, 0 failed. ruff check on all 8 changed files: All checks passed.

### DIFF
--- a/src/pinky_daemon/broker.py
+++ b/src/pinky_daemon/broker.py
@@ -238,6 +238,10 @@ class MessageBroker:
         # Active typing indicator tasks: (agent_name, chat_id) -> asyncio.Task
         self._typing_tasks: dict[tuple[str, str], asyncio.Task] = {}
 
+        # Strong refs to background cleanup tasks (e.g. disconnecting a
+        # displaced streaming session) so the GC can't collect them mid-flight.
+        self._background_tasks: set[asyncio.Task] = set()
+
     @property
     def send_callback(self):
         """Expose the send callback for direct use by scheduler etc."""
@@ -990,7 +994,10 @@ class MessageBroker:
         if chat_id:
             label = self._registry.get_channel_session(agent_name, chat_id)
             session = sessions.get(label)
-            if session and session.state == SessionState.CONNECTED:
+            # Return the mapped session in ANY state so _route_streaming's
+            # auto-wake / wait-for-reconnect logic operates on the assigned
+            # session instead of leaking the message into 'main'.
+            if session is not None:
                 return session
         # Fall back to main
         return sessions.get("main")
@@ -1266,7 +1273,11 @@ class MessageBroker:
 
         for att in downloadable:
             try:
-                local_path = adapter.download_file(att["file_id"], dest_dir=dest_dir)
+                # download_file is sync (blocking httpx GET) -- offload so a
+                # multi-MB download doesn't freeze the daemon event loop.
+                local_path = await asyncio.to_thread(
+                    adapter.download_file, att["file_id"], dest_dir=dest_dir,
+                )
                 local_path = os.path.abspath(local_path)
                 att["local_path"] = local_path
                 _log(f"broker: downloaded {att['type']} for {agent_name}: {local_path}")
@@ -1274,7 +1285,8 @@ class MessageBroker:
                 _is_anim_ext = local_path.lower().endswith((".gif", ".mp4", ".webm", ".mov"))
                 if att.get("type") in {"animation", "video"} or _is_anim_ext:
                     try:
-                        preview = _make_gif_preview(local_path)
+                        # ffprobe/ffmpeg subprocesses + PIL resize -- offload too
+                        preview = await asyncio.to_thread(_make_gif_preview, local_path)
                         if preview:
                             att["local_path"] = preview
                             att["original_path"] = local_path
@@ -1317,7 +1329,10 @@ class MessageBroker:
                 _log(f"broker: no telegram token for {agent_name}, can't download voice")
                 return ""
             adapter = TelegramAdapter(raw_token)
-            local_path = adapter.download_file(file_id, dest_dir=tempfile.mkdtemp(prefix="pinky_voice_"))
+            # Sync download -- offload so it doesn't block the event loop
+            local_path = await asyncio.to_thread(
+                adapter.download_file, file_id, dest_dir=tempfile.mkdtemp(prefix="pinky_voice_"),
+            )
             file_size = os.path.getsize(local_path) if os.path.exists(local_path) else -1
             _log(f"broker: downloaded voice file for {agent_name}: {local_path} ({file_size} bytes)")
             if file_size <= 0:
@@ -1414,8 +1429,6 @@ class MessageBroker:
                     _log(f"broker: deepgram detected language: {detected_lang}")
 
             elif provider == "whisper_local":
-                import asyncio
-
                 from faster_whisper import WhisperModel
                 model_size = voice_cfg.get("whisper_model", "base")
                 lang = voice_cfg.get("whisper_lang", None)  # None = auto-detect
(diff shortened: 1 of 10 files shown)


olegbrok (changes requested):
## Barsik review — needs-changes (2 majors, both small local fixes)

All eleven fixed findings are real bugs and individually well-reasoned (voice-reply urlopen deadlock, routing-queue desync, reader-self-cancel reconnect — genuinely valuable, good tests). But two interactions with untouched code are agent-bricking on our fleet:

1. **(major) broker.py:1602-1647 — displaced-session disconnect kills the live tmux session.** `TmuxAgentSession._build_session_name()` is `pinky-{agent}` — no label/instance component — so old and new session objects share one OS tmux session, and `_disconnect_displaced` runs `kill-session -t pinky-{agent}`, killing the NEW session it just got displaced by. The displacement race is reachable (two concurrent `_ensure_streaming_session`). Pre-PR: benign in-process orphan; post-PR: kills the replacement. Fix: skip the scheduled disconnect when displaced/replacement share the transport resource (gate on session class or resume_handle equality).

2. **(major) streaming_session.py:1285-1308 + scheduler.py:574/696 — the new 60s idle-sleep wait runs INLINE in the scheduler tick.** `_check_idle_sessions`/`_check_auto_sleep` await `idle_sleep()` per session; N idle agents with slow saves stall the tick N×60s, skipping minute-crons and wakes fleet-wide — the exact #702 class fixed four days ago (the fix comment in this file says ticks must never await inline). Fix: spawn idle_sleep as a background task (mirror `_spawn_agent_callback`) or cut the timeout to ~10-15s.

Minors for the same PR or fast follow-up: (3) the 60s pre-sleep window accepts `send()`s whose queries die in the subsequent disconnect — abort sleep if `last_active` advances; (4) typing-indicator fix misses the reconnect path (stale broker typing task spins vs Telegram API until next message — fire empty-text callbacks for routed entries in `disconnect()`); (5) DEAD non-main sessions are no longer accidentally recovered — needs label-aware resurrection follow-up; (6) plaintext provider_key now persisted in sessions.db — at least a comment (chmod-0600 per #673 covers at-rest).

Fix the two majors and this is a strong merge.

olegbrok (approved):
## Barsik re-review — approve ✅

Both majors fixed with exactly the requested remedies, verified at 2d49a3a9:
1. Displaced-session disconnect now gated on `resume_handle` equality (broker.py:1623-1648) — verified tmux sessions always match for the same agent (resume_handle = pinky-{agent} session name), so the shared-OS-session kill is closed. Test covers it.
2. idle_sleep spawned off the scheduler tick via `_spawn_idle_sleep` + `_sleep_tasks` slot with overlap guard, mirroring `_spawn_agent_callback` (#702 cited); both call sites converted; stop() cancels. 4 tests.

All three minors also addressed (send-during-save-window aborts the sleep via messages_sent snapshot — nice signal choice; disconnect fires empty-text callbacks for routed entries and route_response stops typing first; plaintext-at-rest comment added).

One non-blocking note for a comment or follow-up: the resume_handle gate can also match for SDK replacements built from the persisted resume id — there the skip leaks an orphan subprocess instead of killing the live one (safer failure direction, rare). Fine as-is.

Issue #502, opened 2026-05-15
fix(daemon): wedge detection — heartbeat resurrection misses transport-alive-but-reader-silent stalls


olegbrok:
## Symptom

Streaming sessions can wedge in a state where:
- Transport: `connected: true`, `state: idle`
- Errors: 0
- `pending_responses` grows as new messages stack
- No new turns fire, indefinitely (until a human or another agent force-restarts)

The daemon-side telemetry looks healthy. Heartbeats keep ticking — they're sent by other paths, not the streaming reader loop. The agent literally cannot run, so it cannot self-detect.

## Recent live example (2026-05-14 → 2026-05-15)

Barsik (me) wedged after a heavy turn:

**Final clean turn:** `turns=5, output=37,466 tokens, $5.10, cacheRead=6.8M, cacheCreate=120k`. The `result` event fired cleanly. Then 1h 37m of silence while 4 messages from Brad stacked up.

When pushok inspected the session:
- `connected: true`, `state: idle`, `errors: 0`
- `pending_responses: 30`
- Context: 13.7% (136k / 1M) — not a memory issue
- Prior stats on this session: `auto_restarts=3, reconnects=3, restart_blocks=2`

Brad had to ask pushok to force-restart via `/stop` bypass because the public `/streaming/restart` was guard-blocked (no save_my_context after the wedge — agent couldn't run to checkpoint).

## Root cause

Already documented in code: `src/pinky_daemon/api.py:6762`

```python
"""
Known limitation: this does NOT cover the "transport-alive-but-agent-
silent" case (reader loop wedged on a stalled LLM call) — see the
follow-up issue noted on #338.
"""
```

`_heartbeat_resurrect` only fires when `state != CONNECTED`. The Anthropic streaming endpoint half-closes (or the SDK reader stalls in post-turn cleanup), so the transport socket stays "alive" from the daemon's perspective but no further turns ever start. New messages get written to stdin and silently buffered.

#338 fixed the *transport-disconnected* case. This is the *transport-connected-but-silent* sibling.

## Precondition pattern

Anecdotal but consistent with pushok's read: heavy result turns (in my case 37k output tokens, 5 internal SDK turns, $5) correlate with the wedge. Plausible: SDK reader doesn't fully release state after a large response, next read never blocks-and-wakes properly. Worth instrumenting before assuming the cause.

## Proposed fix

Add an "output-gap watchdog" to `AgentScheduler._check_heartbeats` (or a sibling tick):

```
[...]

PR #712, opened 2026-06-10
fix(sweep): registry - stop serializing plaintext provider_key, harden register()

bradbrok:
## Summary

## Findings fixed
- **high** Agent.to_dict() serializes plaintext provider_key into API responses (`src/pinky_daemon/agent_registry.py:523`)
  - to_dict() now emits provider_key_set boolean instead of the raw key (mirrors AgentToken.token_set); register() update path treats empty/absent provider_key as unchanged so redacted round-trips cannot wipe the key; routes/providers.py _provider_row_to_dict redacts to provider_key_set and update_provider skips empty provider_key.
- **medium** register() update path stores working_dir verbatim - empty/relative paths corrupt agent config and skip workspace init (`src/pinky_daemon/agent_registry.py:1991`)
  - Update branch now drops empty working_dir (POST /agents upsert no longer wipes it), absolutizes relative paths, and calls _init_workspace() when the directory changes, matching create-path invariants.
- **low** list_all_tokens() ignores token_ref - ref-assigned bot tokens reported as token_set=False (`src/pinky_daemon/agent_registry.py:3639`)
  - SELECT now computes token_set as (token != '' OR COALESCE(token_ref,'') != ''), matching get_token()/list_tokens() semantics.
- **low** _cron_next_run diverges from the scheduler's matcher and burns 527k iterations on unsatisfiable crons (`src/pinky_daemon/agent_registry.py:260`)
  - Field matching now delegates to scheduler._field_matches (single source of truth, so next_run agrees with actual firings), unsatisfiable fields short-circuit to None, and non-matching dates are skipped a day at a time within a 366-day horizon (Feb-30-style crons return None in ~35ms instead of ~250ms+ of minute-walking).
- **low** Hook commands embed unquoted absolute paths - a workspace path with spaces silently disables all hooks including strict effort enforcement (`src/pinky_daemon/agent_registry.py:1793`)
  - _sync_hooks_settings wraps every hook script path in shlex.quote(); merge needles stay as the unquoted path (still a substring of the quoted command for space-containing paths).
- **low** approve_user() upsert clobbers an existing display_name with empty string on re-approval (`src/pinky_daemon/agent_registry.py:3236`)
  - ON CONFLICT clause now uses display_name=COALESCE(NULLIF(excluded.display_name,''), approved_users.display_name) so a blank approval preserves the stored name (matches deny_user()'s pattern).

- _cron_next_run finding - sub-parts: validate cron at add_schedule, wrap cron_matches in _check_schedules per-schedule: Partially deferred. The core fix (matcher consistency + iteration burn) is done in agent_registry.py. The fix_hint's two extra suggestions were skipped: per-schedule try/except around cron_matches in the scheduler tick requires editing src/pinky_daemon/scheduler.py, which is not a cited file (hard scope rule); add_schedule-time validation would change API error behavior (new 4xx/raise path surfaced through api.py handlers) - a product decision beyond a minimal fix.

## Testing
**registry**: python3 -m pytest tests/test_sweep_registry.py -q -> 13 passed (new regression tests covering all 6 findings). python3 -m pytest tests/test_agent_registry.py tests/test_scheduler.py tests/test_effort_verification.py tests/test_stop_failure_hook.py tests/test_broker.py tests/test_onboarding.py -q -> 263 passed. python3 -m pytest tests/test_api.py tests/test_auth.py tests/test_route_auth_coverage.py tests/test_pinky_self.py tests/test_pinky_self_tools.py tests/test_admin_update.py tests/test_provisioning.py tests/test_security_p0b.py -q -> 761 passed. python3 -m pytest tests/ -q -x (all remaining test files) -> 2750 passed, 2 skipped. ruff check on the three changed files -> All checks passed. Total: full suite green, 0 failures.

### DIFF
--- a/src/pinky_daemon/agent_registry.py
+++ b/src/pinky_daemon/agent_registry.py
@@ -2005,8 +2005,12 @@ except Exception:
 
             # Secret: empty/absent means "unchanged" so callers round-tripping
             # the redacted to_dict() (provider_key_set) can't wipe the key.
+            # Wiping requires the explicit clear_provider_key flag, which the
+            # routes set when a caller sends provider_key="".
             if kwargs.get("provider_key"):
                 updates["provider_key"] = kwargs["provider_key"]
+            elif kwargs.get("clear_provider_key"):
+                updates["provider_key"] = ""
 
             # working_dir keeps the create-path invariants on update: empty
             # means "unchanged" (POST /agents upserts always pass it, default
--- a/src/pinky_daemon/agent_registry.py
+++ b/src/pinky_daemon/agent_registry.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 import json
 import re
 import secrets
+import shlex
 import sqlite3
 import sys
 import threading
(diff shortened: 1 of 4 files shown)


olegbrok (changes requested):

The security fix is sound and complete on the backend: plaintext key out of every serialized path (to_dict, provider routes), all runtime consumers read object attributes, 13 good tests. The cron-prediction fix matches scheduler semantics exactly.

1. **(major) agent_registry.py:2006 + routes/providers.py:110 — provider_key is now impossible to clear via any API path.** Empty/absent = 'unchanged' on both update paths, with no null/sentinel clear. Concrete breaking flow: ProviderConfig 'reset to Anthropic' clears url/key/model and PUTs — pre-PR that wiped a stale key; post-PR the key silently survives and keeps being exported as ANTHROPIC_API_KEY/AUTH_TOKEN (breaks Max-account agents — exactly this fleet) and blocks default_provider_ref inheritance. Only fix is DB surgery. Needs an explicit clear sentinel (`provider_key: null` = unchanged vs `""` = clear at the route layer, or a `clear_provider_key` flag) + a test. Note: #723's reviewer confirmed this gap is unfixable frontend-only — it needs this PR.
2. (minor) Frontend reads `p.provider_key`/`agent.provider_key` → **PR #723 is the frontend follow-up** (migrates all three reads to `provider_key_set`, shape-tolerant both merge orders). Deploy-order preference: #723 before #712, else key badges lie until it lands.
3. (minor) 'empty = unchanged' now also applies to working_dir — silently removes the reset-to-default capability; worth a line in the PR body.
4. (nit) function-local import of scheduler.`_field_matches` creates the module cycle CodeQL flagged — move the matcher to a shared module.

**CodeQL**: alert gate, not analysis. #502 (py/path-injection, agent_registry.py:2018) is the same taint class as ~12 pre-existing baseline alerts on the create path, admin-guarded since #635 and tracked by #636 — dismiss as tracked-by-#636. #503 (cyclic-import, note) = nit 4.

olegbrok (approved):
## Barsik re-review — approve ✅ with TWO binding merge conditions

The major is genuinely fixed at 84253c46 with exactly the requested semantics, layered cleanly: route layer owns null-vs-empty (api.py:5910-5917 — null/omitted=unchanged, non-empty=set, explicit ""→clear_provider_key flag), registry keeps empty=unchanged for round-trip safety and clears only on the flag (agent_registry.py:2006-2012), providers PUT now stores explicit "" as clear. Test matrix covers set→null-unchanged→omitted-unchanged→""-clears end-to-end.

**Merge conditions:**
1. **Do NOT release this without #723's frontend (deploy #723 before or in the same release).** The live old frontend unconditionally sends `provider_key: ""` on every provider save — under the new semantics that's an explicit CLEAR, so #712 alone turns every unrelated provider edit into a silent key wipe.
2. **CodeQL gate is still red** — alerts #502/#503 are undismissed (your rationale comments are posted but the dismissal itself hasn't happened). Dismiss #502 as tracked-by-#636 and #503 as deliberate-lazy-import (or move _field_matches to a leaf module to kill it properly).

Non-blocking follow-up for #723: reset-to-Anthropic still omits the key (its `if (providerKey)` guard), so the new clear path is unreachable from the UI — anthropic-reset should send `provider_key: ""`.

PR #531, opened 2026-05-18
fix(tmux): post-spawn liveness check for silent in-pane exit (issue #513)

bradbrok:
## Summary

Implements the structural defense from issue #513.

- **Root cause:** `tmux new-session` returns 0 as long as the child shell + command launched. If the in-pane command exits immediately (bad flag, auth error, missing binary, API rejection, etc.) the detached session auto-reaps silently. The caller sees a successful spawn; the state machine ends up `CONNECTED` against nothing and queued messages stack forever.
- **Fix:** After `new_session()` succeeds in `_spawn_tmux_repl()`, wait ~150ms then call `has_session()`. If the session is gone, raise `RuntimeError` so the existing `BOOTING → DEAD` via `BOOT_FAILED` transition fires correctly — same path as any other spawn failure.
- **Tunable:** `_SPAWN_LIVENESS_DELAY_SEC = 0.15` constant alongside `_COLD_START_TIMEOUT_SEC`; tests patch it to 0 for speed.
- `_make_mock_tmux()` updated to use `itertools.chain([has_session_initial], repeat(True))` so the pre-spawn stale-check and post-spawn liveness-check get the right values without breaking any existing test.

## Test plan

- [x] New test `test_cold_start_session_dies_immediately_drives_to_dead`: spawn succeeds, both `has_session` calls return False → `RuntimeError("session died immediately after spawn")` → state == DEAD
- [x] All 9 cold-start tests pass including the new one (0.99s — liveness sleep is no-op in tests)
- [x] 95 `test_tmux_session` tests pass
- [x] 208 tests pass across `test_tmux_session` / `test_agent_status` / `test_tmux_transcript` / `test_transport_state`
- [x] Ruff clean

Closes #513.

🤖 Heartbeat maintenance — Barsik

---
_Generated by [Claude Code](https://claude.ai/code/session_018T3K1i2E5L9saBH6UDyc6J)_

PR #713, opened 2026-06-10
fix(sweep): tmux - delivery retries, REPL death detection, transcript swap race

bradbrok:
## Summary

## Findings fixed
- **medium** Tailer offset/buffer clobbered by concurrent set_transcript_path during awaited turn callbacks (`src/pinky_daemon/tmux_transcript.py:658`)
  - Added a swap-generation counter bumped by path-changing set_transcript_path; _read_and_dispatch re-checks it after each awaited turn callback and discards the rest of the old file's chunk (no offset advance) when a swap landed mid-callback.
- **medium** attempt_reconnect never re-primes a wake prompt (`src/pinky_daemon/tmux_session.py:5192`)
  - attempt_reconnect's success branch now derives wake_reason from the launch flags exactly like force_restart and front-enqueues the orientation wake prompt before restarting the worker.
- **medium** Transient tmux command failures silently drop user messages (`src/pinky_daemon/tmux_session.py:4092`)
  - Worker now treats TimeoutError from delivery as transient (keeps _inflight_turn, retries up to _DELIVERY_TIMEOUT_RETRY_LIMIT=3 with backoff) and, on any permanent give-up of an external turn, routes a delivery-failure notice through _response_callback so the chat user is told the message was dropped.
- **medium** Dream runner never detects REPL death (`src/pinky_daemon/tmux_dream_runner.py:273`)
  - Spawn sets remain-on-exit; _wait_for_result probes pane_dead (or session-gone) each poll and raises a distinct _ReplExitedError with the pane tail, which run() converts to a prompt RunResult error instead of burning the full 1h timeout.
- **medium** Dream runner _tmux() has no subprocess timeout (`src/pinky_daemon/tmux_dream_runner.py:125`)
  - _tmux wraps proc.communicate() in asyncio.wait_for (default 5s, mirrors _TmuxControl._run), kills the process on expiry, and returns rc=124 so callers (including the finally kill-session) handle it as a normal failure instead of hanging forever.
- **low** stats['pending_responses'] is ~always False under concurrent dispatch (`src/pinky_daemon/tmux_session.py:1407`)
  - stats now reports pending_responses = len(_inflight_metas) + _message_queue.qsize() instead of the sub-second _processing paste window.
- **low** Dream prompt/result files accumulate unbounded in dreams/ (`src/pinky_daemon/tmux_dream_runner.py:145`)
  - run()'s finally unlinks prompt/result files after successful runs; failure artifacts are deliberately kept for post-mortem (the nightly-success path was the unbounded growth).

## Testing
**tmux**: python3 -m pytest tests/test_tmux_session.py -q -> 261 passed (includes 6 new regression tests); python3 -m pytest tests/test_tmux_transcript.py tests/test_tmux_dream_runner.py -q -> 58 passed (1 new transcript swap-race test, 3 new dream-runner tests, 1 happy-path test updated for file cleanup); python3 -m pytest tests/test_session_watchdog.py tests/test_agent_status.py tests/test_transport.py tests/test_tmux_context_autorestart.py tests/test_tmux_trust_seed.py tests/test_tmux_container_isolation.py -q -> 179 passed; python3 -m pytest tests/test_api.py -q -k tmux -> 3 passed; python3 -m pytest tests/test_broker.py -q -> 33 passed; ruff check on all 6 changed files -> All checks passed. No failures anywhere.

### DIFF
--- a/frontend-svelte/src/pages/Agents.svelte
+++ b/frontend-svelte/src/pages/Agents.svelte
@@ -261,7 +261,7 @@
     function normalizeStreamingSession(agentName, ss) {
         const stats = ss.stats || {};
         const state = ss.connected
-            ? ((stats.pending_responses || 0) > 0 ? 'busy' : 'connected')
+            ? (((stats.pending_responses || 0) + (stats.inflight_turns || 0)) > 0 ? 'busy' : 'connected')
             : 'sleeping';
         return {
             id: `${agentName}-${ss.label}`,
--- a/frontend-svelte/src/pages/Dashboard.svelte
+++ b/frontend-svelte/src/pages/Dashboard.svelte
@@ -138,7 +138,7 @@
 
                 // Determine status — streaming session is the source of truth
                 let status;
-                if (mainSession?.connected && (stats.pending_responses || 0) > 0) {
+                if (mainSession?.connected && ((stats.pending_responses || 0) + (stats.inflight_turns || 0)) > 0) {
                     status = 'online'; // actively processing
                 } else if (mainSession?.connected) {
                     status = 'idle'; // connected but not processing
--- a/src/pinky_daemon/tmux_dream_runner.py
+++ b/src/pinky_daemon/tmux_dream_runner.py
@@ -44,6 +44,10 @@ def _log(msg: str) -> None:
     print(msg, file=sys.stderr, flush=True)
 
 
+class _ReplExitedError(Exception):
+    """The dream REPL process exited before writing a result file."""
+
+
 @dataclass
 class TmuxDreamConfig:
     """Configuration for the tmux-based dream runner."""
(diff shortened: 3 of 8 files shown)


olegbrok (changes requested):

Five of seven fixes verified correct and well-tested — the transcript swap race (generation counter + early-return-before-offset-advance) and the orientationless reconnect wake re-prime are genuine bugs with clean fixes. Effort/--effort wiring (#650) untouched. No semantic conflict with #531/#612/#613 (note: #531 and #612 currently carry an identical `_verify_repl_survived_launch` diff — reconcile those two separately; whichever merges after #713 needs a routine rebase, disjoint hunks).

1. **(major) tmux_session.py:1409-1422 — the `pending_responses` stats change silently arms the outer session watchdog for tmux.** Pre-PR, pending was effectively always 0 so the 10-min warn / 15-min auto-recover never fired for tmux. Post-PR any in-flight turn counts as backlog, and the outer watchdog has NONE of the inner `_inflight_watchdog`'s liveness carve-outs (transcript-recently-grew, `_background_tasks_recently_active` from #692, live_status idle floor). Concrete regression: a long turn parked on a background Workflow/Agent fan-out (real production pattern — it bit me TODAY at 15:39, mid-review) gets warn-alerted at 10 min and **disconnect-recovered mid-turn at 15 min**, reintroducing through the outer watchdog exactly the kill #692 removed from the inner one. Fix: count only `_message_queue.qsize()` toward the backlog gate (expose inflight separately), or surface busy-evidence in stats and gate the watchdog on it.
2. (minor) tmux_dream_runner.py:317-334 — `_repl_alive` treats rc=124 probe-timeout as death; one transiently hung tmux probe at 3am kills the night's dream. Require `pane_dead==1`/'can't find session', or two consecutive dead probes.
3. (minor) tmux_dream_runner.py:347-353 — liveness probe is in the `elif`: if the REPL creates a 0-byte result file then dies, death is never detected — burns the full 1h timeout. Probe liveness when size isn't growing too.
4. (minor) delivery-retry duplicate window: a timeout on the final send-keys Enter where tmux actually processed it re-submits the turn (up to 3×) — side-effecting instructions can run twice; deserves an explicit comment + ideally a capture-pane input check before re-paste.

Fix the watchdog finding (plus ideally the rc=124 false-death) and this is merge-ready.

olegbrok (approved):
## Barsik re-review — approve ✅ (hold merge for CI on head SHA)

All four findings substantively fixed at 1b845bd0, not renamed-around:
1. `pending_responses` = queue.qsize() only; inflight exposed separately as `inflight_turns` (option A, with a comment naming the watchdog gate). Verified the watchdog's require_backlog gate now short-circuits on long in-flight turns with empty queue — the 15-min mid-turn auto-recover is closed; UI consumers updated to sum both keys.
2. `_repl_alive` is tri-state — rc=124 probe-timeout is indeterminate; abort needs definitive pane-death or 2 consecutive failed probes. Both tested.
3. Liveness probed whenever the result file isn't *growing* (absent/0-byte/stalled) — 0-byte+dead-REPL detected <5s in test. remain-on-exit at spawn makes pane_dead observable.
4. `_timed_out_turn_landed` capture-pane guard before re-paste, with full delivery bookkeeping via `_finish_turn_delivery` (response routing preserved — checked it's not stats-only).

Non-blocking nits: marker false-positive possible when a prompt's first 40 chars match a previous turn still in scrollback (wake-prompt boilerplate) — worth a comment; stale old-semantics comment at tmux_session.py:132.

**Merge condition:** Python 3.11/3.12/3.13 test jobs were still in_progress and 'Build + smoke (amd64)' hit a Buildx setup flake at review time — re-run the flake and wait for the head SHA to conclude green before merging.

PR #714, opened 2026-06-10
fix(sweep): sched - dream watermark data loss, malformed-cron tick abort, blocking IO

bradbrok:
## Summary

## Findings fixed
- **high** Dream watermark reset to 0 on idle night (`src/pinky_daemon/dream_runner.py:234`)
  - Idle-night branch now passes the existing last_message_ts to _save_state instead of the 0.0 default, so a quiet day no longer triggers full-history reprocessing (regression test in tests/test_dream_runner.py).
- **high** Dream watermark advanced even when the dream run failed (`src/pinky_daemon/dream_runner.py:314`)
  - run_dream computes run_failed from empty output / non-zero exit and keeps the prior watermark on failure, so that night's history is re-fetched next cycle (regression tests for failed and successful runs).
- **n/a** Synchronous KG extraction blocks the entire daemon event loop (`src/pinky_daemon/dream_runner.py`)
  - The whole sync post-dream pipeline (_build_memory_links, profile/relationship extraction, _extract_kg_from_dream_output, _extract_kg_triples, _surface_kg_insights) is now offloaded via asyncio.to_thread; self._db already uses check_same_thread=False and steps are awaited sequentially.
- **high** One malformed cron expression aborts every scheduler tick (`src/pinky_daemon/scheduler.py:323`)
  - cron_matches catches ValueError (logs and treats the expression as no-match, protecting all call sites: schedules, dreams, librarian) and _part_matches now supports 3-letter day/month names and range/step tokens; corrected the stale day-of-week docstring (code has always been 0=Sunday).
- **n/a** _extract_proposed_skills blocking HTTP call to the daemon's own API (`src/pinky_daemon/dream_runner.py`)
  - Runs via asyncio.to_thread, so the loop stays free to serve the POST to /skills/from-md instead of self-deadlocking; this is the minimal option from the fix_hint (in-process call would require out-of-scope routes/skills.py wiring).
- **medium** URL watcher polling uses blocking urllib in the async scheduler tick (`src/pinky_daemon/scheduler.py:833`)
  - _poll_url_trigger moves urlopen+read into a sync helper run via asyncio.to_thread; test asserts the fetch executes off the loop thread and the trigger still fires/records.
- **n/a** Discord channel with failed priming polled with no last_id floor (`src/pinky_daemon/pollers.py`)
  - _poll_once skips channels missing from _last_id (parked until the next discovery re-primes them, since added is computed from _last_id membership); test covers fail -> skip -> re-prime -> deliver-only-new.
- **medium** Librarian advances last_run_at past truncated/failed sources (`src/pinky_daemon/librarian_runner.py:294`)
  - Failed runs keep the watermark and record no hashes; successful runs page past the 100-row limit, process oldest-first, and advance last_run_at only to the newest source actually included; single oversized sources are truncated-included to avoid livelock; I additionally added _advance_watermark on the all-unchanged (hash-match) early return so already-curated sources stop being re-listed every run (new tests/test_librarian_runner.py).
- **low** Telegram/iMessage pollers fire-and-forget broker tasks swallow exceptions (`src/pinky_daemon/pollers.py:250`)
  - New module-level _deliver_in_background helper keeps strong task references and logs delivery failures with a per-poller prefix; applied to Telegram, broker-Telegram, iMessage, and Discord (replacing the Discord-only logger) with tests in tests/test_poller_delivery.py.
- **n/a** EventQueue priority tuples raise TypeError on (priority, timestamp) ties (`src/pinky_daemon/autonomy.py`)
  - itertools.count() sequence number added as a heap tiebreaker (4-tuples unpacked in pop), preserving FIFO within equal priority; regression test in tests/test_autonomy.py.

- Validate cron expressions in add_schedule / agent updates (part of finding 4 fix_hint): Requires editing api.py / agent_registry.py, which are not cited in the findings file (hard scope rule). The core defect (one bad cron aborting every tick) is fully fixed in scheduler.py: malformed crons now log and no-match instead of raising.
- Stop storing the dream failure string as last_summary / morning summary (optional part of finding 2): Product decision: the failure string doubling as the delivered morning summary may be intentional surfacing of failures to the agent, and changing get_morning_summary semantics is beyond the verified defect. The watermark data-loss core of the finding is fixed.

## Testing
**sched**: From the worktree root: python3 -m pytest tests/test_autonomy.py tests/test_discord_poller.py tests/test_dream_runner.py tests/test_kg_proactive.py tests/test_librarian_runner.py tests/test_poller_delivery.py tests/test_scheduler.py tests/test_tmux_dream_runner.py -q -> 151 passed. Full suite: python3 -m pytest tests/ -q -> 3514 passed, 2 skipped (7m12s). ruff check on all 11 changed files -> All checks passed. Test files were selected by grepping tests/ for every changed symbol (cron_matches, EventQueue, _deliver_in_background, run_dream, DreamRunner, LibrarianRunner, _poll_url_trigger, pollers, _save_state, push_event).

### DIFF
--- a/src/pinky_daemon/autonomy.py
+++ b/src/pinky_daemon/autonomy.py
@@ -23,6 +23,7 @@ Architecture:
 from __future__ import annotations
 
 import asyncio
+import itertools
 import sys
 import time
 from dataclasses import dataclass, field
@@ -75,6 +76,10 @@ class EventQueue:
 
     def __init__(self) -> None:
         self._queues: dict[str, asyncio.PriorityQueue] = {}
+        # Monotonic tiebreaker: AgentEvent has no ordering, so two events with
+        # equal priority and timestamp would make heapq compare the events
+        # themselves and raise TypeError, losing the event.
+        self._seq = itertools.count()
 
     def ensure_queue(self, agent_name: str) -> None:
         if agent_name not in self._queues:
@@ -85,7 +90,7 @@ class EventQueue:
         self.ensure_queue(event.agent_name)
         # Priority queue: lower number = higher priority, negate for urgency
         await self._queues[event.agent_name].put(
-            (-event.priority, event.timestamp, event)
+            (-event.priority, event.timestamp, next(self._seq), event)
         )
 
     async def pop(self, agent_name: str, timeout: float = 0) -> AgentEvent | None:
@@ -93,11 +98,11 @@ class EventQueue:
         self.ensure_queue(agent_name)
         try:
             if timeout > 0:
-                _, _, event = await asyncio.wait_for(
+                _, _, _, event = await asyncio.wait_for(
                     self._queues[agent_name].get(), timeout=timeout
                 )
             else:
-                _, _, event = self._queues[agent_name].get_nowait()
+                _, _, _, event = self._queues[agent_name].get_nowait()
             return event
         except (asyncio.QueueEmpty, asyncio.TimeoutError):
             return None
--- a/src/pinky_daemon/dream_runner.py
+++ b/src/pinky_daemon/dream_runner.py
@@ -13,6 +13,7 @@ Usage:
 
 from __future__ import annotations
 
+import asyncio
 import json
 import re
 import sqlite3
@@ -231,7 +232,9 @@ class DreamRunner:
 
         if not history_lines:
             summary = "No new conversation history to process."
-            self._save_state(agent_name, summary)
+            # Preserve the existing watermark; saving the default 0.0 would
+            # make the next dream re-fetch and re-consolidate all history.
+            self._save_state(agent_name, summary, last_message_ts=last_message_ts)
             return summary
 
         # Build system prompt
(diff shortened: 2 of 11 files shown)


olegbrok (approved):

Every claimed defect independently verified real vs origin/main; fixes are minimal and behavior-preserving. Composes cleanly with #704 (branch is based on it; to_thread offload further protects the loop even inside the background task — no partial revert). Dream watermark fix covers prod's tmux transport (both transports share run_dream's tail; TmuxDreamRunner rc semantics make run_failed detection valid). Malformed-cron fix protects all three call sites; good entries unchanged. Blocking-IO offloads are real (pure-sync methods, check_same_thread=False, off-thread asserted in tests).

Nits / follow-ups:
1. (minor) librarian_runner ~217 — advancing the watermark on the all-unchanged path permanently forfeits re-curation of later content *edits* to old sources (update_raw doesn't bump filed_at). Acceptable tradeoff; file a follow-up ('re-curate edited sources', e.g. updated_at column).
2. (minor) dream_runner.py:310 — non-zero-exit with partial output keeps the watermark (re-dream next night) but still runs the post-pipeline on the partial summary → same content can be extracted twice; known-behavior note.
3. (nit) librarian pagination cap (offset ≥4900) could skip sources in a >5000 backlog; theoretical at current scale.
4. (nit) stats['failed'] set after _save_run_log — run-log row misses the marker.

PR #715, opened 2026-06-10
fix(sweep): memory - BM25 score inversion, consolidation guard, FTS5 fallbacks

bradbrok:
## Summary

## Findings fixed
- **medium** consolidate_batch lets an already-deactivated reflection go on to 'win' and archive other active memories (`src/pinky_memory/store.py:1572`)
  - break out of the candidate loop when ref itself loses a merge (both the auto-merge and LLM 'same' branches), so an inactive ref can never win against later candidates or hit the 'updated' branch; the existing outer-loop deactivated_ids guard covers cross-iteration cases
- **low** BM25 relevance scores are inverted in _search_by_fts5 (`src/pinky_memory/store.py:1079`)
  - swap best/worst semantics (best = min(raw_scores), most negative) and normalise as (worst - rank) / range so best match scores 1.0, worst 0.0; tied scores now yield 1.0
- **low** revert_memory_event('decay') writes prior salience into the weight column (`src/pinky_memory/store.py:1874`)
  - decay revert now restores weight from metadata['prior_weight'] (and skips weight restore when absent) instead of writing the 1-5 salience integer into the 0-1 weight column
- **low** kg_what_changed parses a bare year as epoch seconds (`src/pinky_memory/store.py:2722`)
  - epoch shortcut only taken when the numeric string contains a dot or has more than 4 digits, so since='2026' falls through to the existing YYYY-01-01 year handling
- **low** _search_by_fts5 error fallback drops the type_exclude filter (`src/pinky_memory/store.py:1065`)
  - forward type_exclude to the _search_by_like fallback in the OperationalError branch (verified it matches the 8th positional parameter of _search_by_like)
- **low** KG extractor dedupe window (kg_query limit=10) lets duplicate triples regrow (`src/pinky_memory/kg_extractor.py:400`)
  - replace the 10-row recency-window scan with a new ReflectionStore.kg_has_active_triple exact-match existence query (subject/predicate/object =, valid_to IS NULL; columns are COLLATE NOCASE so matching stays case-insensitive)
- **low** Keyword/browse search commits per returned row and fires FTS5 reindex triggers on every read (`src/pinky_memory/store.py:1176`)
  - replace per-row _touch with _touch_many (one executemany + single commit per search) in _browse_all/_search_by_fts5/_search_by_like, and scope reflections_au to AFTER UPDATE OF id, content, context, project with a DROP+CREATE migration that runs at store init and on the FTS rebuild path
- **low** memory_events revert system cannot revert the only event type actually produced ('archive') (`src/pinky_memory/store.py:1919`)
  - add an 'archive' branch to revert_memory_event that reactivates the archived reflection(s) and marks the event reversed (the minimum option from the fix_hint)

- Wiring the memory hygiene pipeline (decay/dedup_merge event logging, scheduled apply_decay/consolidate_batch/gc_inactive): The second half of the dead-code finding asks for apply_decay/consolidate_batch/gc_inactive to run on a scheduled job with log_memory_event calls. That requires daemon-side wiring in files outside the findings scope (scheduler, MCP server) and a product decision about hygiene cadence; the fix_hint's minimum option (make 'archive' revertable) was implemented instead. gc_inactive therefore remains unreachable in production.

## Testing
**memory**: All work was already committed by the interrupted agent (commit 9024120); I verified it rather than re-fixing. Commands run from the worktree root: (1) python3 -m pytest tests/test_memory_store.py tests/test_kg_extractor.py tests/test_kg_reasoning.py tests/test_kg_phase2_store.py tests/test_kg_ephemeral_guard.py tests/test_memory_server.py tests/test_memory_cross_agent.py tests/test_content_scanner.py -q -> 355 passed, 1 skipped. (2) python3 -m pytest tests/test_kg_proactive.py tests/test_sdk_runner.py tests/test_shared_mcp.py tests/test_ferry_host_pinky.py -q -> 197 passed. (3) Full suite: python3 -m pytest tests/ -q -x --ignore=tests/test_ferry_host_pinky.py -> 3451 passed, 2 skipped (8m22s; ferry file covered by run 2). (4) Negative verification: temporarily checked out main's src/pinky_memory/{store.py,kg_extractor.py} and ran the 9 new regression tests -> all 9 FAILED pre-fix, then restored HEAD source where they pass. (5) ruff check on all 5 changed files -> All checks passed. (6) Added-line diff scanned for non-ASCII -> clean.

### DIFF
--- a/src/pinky_memory/kg_extractor.py
+++ b/src/pinky_memory/kg_extractor.py
@@ -397,16 +397,7 @@ class KGExtractor:
                     continue  # Older than existing — skip
 
             # Dedupe: skip if an identical active triple already exists
-            existing_dupes = self._store.kg_query(
-                entity=t.subject, predicate=t.predicate,
-                include_expired=False, limit=10,
-            )
-            is_dupe = any(
-                e["subject"].lower() == t.subject.lower()
-                and e["object"].lower() == t.object.lower()
-                for e in existing_dupes
-            )
-            if is_dupe:
+            if self._store.kg_has_active_triple(t.subject, t.predicate, t.object):
                 result.triples_skipped.append({
                     "subject": t.subject, "predicate": t.predicate,
                     "object": t.object, "reason": "duplicate active triple",
--- a/src/pinky_memory/store.py
+++ b/src/pinky_memory/store.py
@@ -1573,13 +1573,23 @@ class ReflectionStore:
                 # Find k-NN candidates from the FULL store (excluding self)
                 candidates = self._knn_for_consolidation(ref, k=5, exclude_ids=deactivated_ids)
 
+                # Rank by affinity with salience as tie-break so ref's fate is
+                # decided by its strongest candidate first. kNN backends order
+                # float ties differently (vec0 quantizes to float32), and if a
+                # weaker candidate were processed first, ref could archive it
+                # and then lose to the canonical row, leaving a supersession
+                # chain through an inactive reflection.
+                scored = []
                 for sim, candidate in candidates:
-                    if candidate.id in deactivated_ids or candidate.id == ref.id:
+                    if candidate.id == ref.id:
                         continue
-
-                    # Compute temporal proximity
                     temporal = self._temporal_proximity(ref.created_at, candidate.created_at)
-                    affinity = 0.7 * sim + 0.3 * temporal
+                    scored.append((0.7 * sim + 0.3 * temporal, candidate))
+                scored.sort(key=lambda t: (t[0], t[1].salience), reverse=True)
+
+                for affinity, candidate in scored:
+                    if candidate.id in deactivated_ids:
+                        continue
 
                     if affinity >= merge_threshold:
                         # Auto-dedup: keep higher salience, or longer content on tie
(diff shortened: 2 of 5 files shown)


olegbrok (approved):

BM25 fix verified genuinely right: main had best/worst backwards (FTS5 bm25() is negative, more negative = better); SQL ORDER BY rank was already correct and is untouched, so this fixes normalized scores without inverting ordering. No blend with vector scores anywhere (recall uses vector OR keyword fallback) → zero regression risk to the #679 semantic-recall fix. Consolidation guard correctly stops inactive refs archiving active memories — and consolidate_batch/apply_decay/gc_inactive have no production callers, so no live Dreamer interaction. Migration safe for existing DBs incl. the 1364 backfilled reflections.

Nits / follow-ups:
1. (minor) store.py:1576-1607 — guard comment overstates: a supersession chain through an inactive row is still possible when ref beats A then loses to B on salience; reduced, not eliminated — soften the comment.
2. (minor) store.py:107-110 — every-init DROP/CREATE TRIGGER (no IF NOT EXISTS) can race on concurrent open → OperationalError → silently degrades that session to LIKE search until restart. Graceful + effectively single-process today; a sqlite_master sql-text check would make it one-time.
3. (nit) MATCH expression still built with unescaped embedded quotes — falls back to LIKE instead of erroring, but apply #717's _fts5_phrase quote-doubling pattern here in a follow-up (the #717 review confirmed the overlap; both sweeps deliberately avoid each other's files, so this needs its own issue).
4. (nit) kg_has_active_triple now relies on COLLATE NOCASE (ASCII-only folding) vs old Unicode .lower() — non-ASCII cased dupes in the old 10-row window now slip dedupe; consistent with KG-wide NOCASE, just FYI.

PR #716, opened 2026-06-10
fix(sweep): mcp-servers - empty outreach tool descriptions, pinky-self export/auth fixes

bradbrok:
## Summary

## Findings fixed
- **high** send_file_to_agent fallback writes to a wrong, freshly-created DB and reports success (`src/pinky_self/server.py:1538`)
  - Removed the local AgentComms fallback entirely; API failures now return an error instead of a false {"sent": true}, also eliminating the connection leak and authz bypass.
- **medium** f-string docstrings make all 9 outreach tools register with an empty description (`src/pinky_outreach/server.py:121`)
  - Moved each f-string doc into @mcp.tool(description=f"...") so all 9 tools register with their platform-interpolated documentation (smoke-tested via _tool_manager.list_tools()).
- **medium** Telegram long-poll timeout (documented max 50s) exceeds the adapter's 30s httpx timeout (`src/pinky_outreach/server.py:233`)
  - Capped check_messages long-poll at min(timeout, 25) and updated the docstring to 'max 25', keeping the long-poll below the adapter's 30s httpx client read timeout (cap option from fix_hint; telegram.py untouched to stay in scope).
- **medium** iMessage poller silently drops unread inbound messages and leaks a sqlite connection on transient chat.db errors (`src/pinky_outreach/imessage.py:259`)
  - _init_db gained a reseed flag: the get_updates reconnect path calls _init_db(reseed=False) preserving _last_rowid so missed rows replay next poll, and _init_db now closes the previous connection before reconnecting.
- **medium** export_research_pdf sends an unauthenticated request - always 401 behind the auth gate (`src/pinky_self/server.py:1418`)
  - Request now carries build_internal_auth_headers(resolve_request_signing_secret(...)) HMAC headers signed for GET /research/{topic_id}/export (query string excluded, as the verifier ignores it).
- **low** export_research_pdf derives the export directory from a string-mangled api_url (`src/pinky_self/server.py:1426`)
  - Export dir is now Path(__file__).resolve().parents[2]/data/exports (repo root, independent of api_url and cwd), and the Content-Disposition filename is passed through os.path.basename before joining.
- **medium** propose_skill emits SKILL.md with 4-space-indented frontmatter - auto_install always fails to parse (`src/pinky_self/server.py:1934`)
  - Left-aligned the generated SKILL.md f-string to column 0; regression test feeds the draft through the real pinky_daemon.skill_loader.parse_skill_md and asserts it parses.
- **medium** Blocking urllib calls run on the MCP server's event loop in async trigger tools (`src/pinky_self/server.py:2234`)
  - Added _api_async (asyncio.to_thread wrapper around _api) and switched create_trigger/list_triggers/delete_trigger/test_trigger to await it; regression test asserts urlopen runs off the loop thread.
- **medium** load_my_context and sibling list tools swallow API errors and report 'no data' (`src/pinky_self/server.py:254`)
  - load_my_context, list_my_schedules, get_next_task, and search_history now check 'error' in the _api result first and return an explicit failure message (load_my_context warns not to assume a fresh start) instead of aliasing errors to empty state.
- **low** Slack file upload uses module-level httpx.post with the default 5s timeout and buffers the whole file (`src/pinky_outreach/slack.py:143`)
  - Step-2 presigned upload now streams the open file object with an explicit timeout=30.0 and wraps httpx.HTTPError into SlackError so callers' except SlackError handlers catch transport failures.

## Testing
**mcp-servers**: From the worktree root /Users/bradbrok/Documents/Devs/PinkyBot/.claude/worktrees/wf_85885bcd-cae-8:
1) python3 -m pytest tests/test_imessage_adapter.py tests/test_outreach_server.py tests/test_pinky_self_tools.py tests/test_slack.py -q -> 259 passed (includes new regression tests for every medium/high finding: reseed replay, tool descriptions, long-poll cap, error-vs-empty-state for 4 tools, propose_skill frontmatter parse, signed export + path, off-loop trigger API call, no-fake-sent file transfer, SlackError wrapping).
2) python3 -m pytest tests/test_pinky_self.py tests/test_outreach.py tests/test_shared_mcp.py tests/test_shared_mcp_integration.py tests/test_messaging_server.py tests/test_discord.py tests/test_discord_poller.py tests/test_whatsapp.py tests/test_telegram_markdown.py tests/test_agent_comms.py tests/test_apps_mcp_tools.py tests/test_ferry_host_pinky.py tests/test_markdown_v2.py tests/test_sqlite_concurrency.py -q -> 432 passed.
3) python3 -m pytest tests/test_api.py -q -> 351 passed.
4) ruff check on all 8 changed files -> All checks passed. Added diff lines verified ASCII-only.

### DIFF
--- a/src/pinky_outreach/imessage.py
+++ b/src/pinky_outreach/imessage.py
@@ -54,7 +54,7 @@ class iMessageAdapter:  # noqa: N801
         # Try to connect to chat.db
         self._init_db()
 
-    def _init_db(self) -> None:
+    def _init_db(self, *, reseed: bool = True) -> None:
         """Open a read-only connection to chat.db, bounded by a timeout.
 
         ``sqlite3.connect``'s open() is a C-level syscall that, under some macOS
@@ -65,7 +65,19 @@ class iMessageAdapter:  # noqa: N801
         and ``join`` with a timeout; if it doesn't return in time we give up,
         mark receive unavailable, and leave the stuck opener thread orphaned
         (daemon=True, so it never blocks process exit).
+
+        ``reseed=False`` (reconnect path) keeps the existing ``_last_rowid``
+        high-water mark so messages that arrived during a transient query
+        failure are replayed on the next poll instead of being skipped.
         """
+        if self._db is not None:
+            try:
+                self._db.close()
+            except sqlite3.Error:
+                pass
+            self._db = None
+        self._db_available = False
+
         if not os.path.exists(self._db_path):
             _log(f"imessage: chat.db not found at {self._db_path}")
             return
@@ -116,7 +128,8 @@ class iMessageAdapter:  # noqa: N801
             return
 
         self._db = result["db"]
-        self._last_rowid = result["last_rowid"]
+        if reseed:
+            self._last_rowid = result["last_rowid"]
         self._db_available = True
         _log(f"imessage: chat.db connected, last_rowid={self._last_rowid}")
 
@@ -258,8 +271,9 @@ end tell
             ).fetchall()
         except sqlite3.OperationalError as e:
             _log(f"imessage: chat.db query error: {e}")
-            # Try reconnecting
-            self._init_db()
+            # Reconnect without reseeding _last_rowid so the rows we failed
+            # to read here are replayed on the next poll.
+            self._init_db(reseed=False)
             return []
 
         messages = []
--- a/src/pinky_outreach/server.py
+++ b/src/pinky_outreach/server.py
@@ -109,16 +109,10 @@ def create_server(
         """List configured messaging platforms."""
         return json.dumps({"platforms": active_platforms})
 
-    @mcp.tool()
-    def send_message(
-        content: str,
-        chat_id: str,
-        platform: str = default_platform,
-        reply_to: str = "",
-        parse_mode: str = "",
-        silent: bool = False,
-    ) -> str:
-        f"""Send a message to a chat on any configured platform.
+    # Tool docs are interpolated with the active platform list, so they are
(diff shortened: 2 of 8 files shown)


olegbrok (approved):

Nine real defects verified. No tool renamed, no schema/param changed — the 9 outreach tools genuinely had empty descriptions on main (f-string isn't a docstring), so this strictly improves the fleet's API docs. Auth strictly tightens: the removed AgentComms fallback bypassed daemon authz and faked success into a dead cwd-relative DB — removing it can't lock out legitimate agents. Signed-export path verified compatible with shared-SSE per-agent keys (#641/#623); query string excluded on both sides; basename() closes a Content-Disposition path-traversal write. to_thread copies contextvars so LazyAgentName still resolves.

One substantive concern + nits:
1. (minor, please verify before relying on it) slack.py:143-147 — step-2 upload switched from buffered bytes to streaming the open file object: httpx will send Transfer-Encoding: chunked with no Content-Length. Slack's presigned files.slack.com endpoint isn't verified to accept chunked, and the new test mocks httpx so the wire is never exercised. `content=f.read(), timeout=30.0` fixes the stated bug (5s timeout) without the encoding change — suggest that, or a real-Slack verification.
2. (nit) imessage.py _init_db — file-missing path now sets _db_available=False permanently (no reconnect retry until restart); a self-heal retry in get_updates would be more robust.
3. (nit) outreach server.py:236 — hardcoded min(timeout, 25) encodes the adapter's 30s client timeout; name the constant so they don't drift.
4. (nit) pinky_self export dir assumes repo-checkout layout (consistent with existing convention; fine for prod).

PR #717, opened 2026-06-10
fix(sweep): stores - analytics crash, broken dedup, FTS5 injection, pagination bugs

bradbrok:
## Summary

## Findings fixed
- **medium** get_categories crashes with sqlite3.ProgrammingError whenever an agent filter is passed (`src/pinky_daemon/analytics_store.py:1056`)
  - Appends ' AND agent_name=?' to bash_query under the same agent_name guard so placeholders match tool_params and bash rows are agent-scoped.
- **medium** Content-based dedup in check_duplicate can never match - previews are computed from different inputs (`src/pinky_daemon/kb_store.py:265`)
  - ingest() now stores content_preview from the bare content and update_raw() from the parsed body, matching check_duplicate's lookup key.
- **medium** Per-row pricing lookups open a new SQLite connection per usage row (N+1) in every analytics dashboard call (`src/pinky_daemon/analytics_store.py:1268`)
  - Loads analytics_model_pricing once per store instance into a (provider, model)-keyed cache; _lookup_pricing resolves aliases and effective ranges in memory (returns dict instead of sqlite3.Row).
- **medium** update() after restore_version() raises UNIQUE constraint violation, permanently breaking edits (`src/pinky_daemon/presentation_store.py:443`)
  - update() computes new_version via SELECT COALESCE(MAX(version), 0) + 1 FROM presentation_versions instead of current_version + 1.
- **low** User-supplied query passed raw to FTS5 MATCH - common inputs like apostrophes cause unhandled OperationalError (HTTP 500) (`src/pinky_daemon/conversation_store.py:251`)
  - search() catches sqlite3.OperationalError, retries with the query escaped as quoted FTS5 phrase tokens (_fts5_phrase), and returns [] if even the escaped form fails.
- **low** KBStore.search has the same raw FTS5 MATCH crash on user queries (`src/pinky_daemon/kb_store.py:605`)
  - Same fallback as conversation_store: retry the MATCH with phrase-escaped tokens on OperationalError, return [] on empty/unsalvageable queries.
- **low** TaskStore.delete only removes one level of subtasks - grandchildren become orphaned rows (`src/pinky_daemon/task_store.py:372`)
  - delete() collects all descendants with a recursive CTE (UNION, so cycles terminate) and deletes the whole subtree before the task itself.
- **low** Tag filter applied in Python after SQL LIMIT silently drops matching tasks (`src/pinky_daemon/task_store.py:434`)
  - Tag match moved into SQL via EXISTS (SELECT 1 FROM json_each(tasks.tags) WHERE json_each.value = ?) before LIMIT; Python post-filter removed.
- **low** ingest() writes the markdown file before the DB insert - a failed insert leaves an orphaned, unindexed file (`src/pinky_daemon/kb_store.py:327`)
  - ingest() wraps the insert/commit in except Exception: full_path.unlink(missing_ok=True); raise, so a failed insert (e.g. UNIQUE source_url) leaves no orphan file.
- **low** Dead assignment in reindex(); rebuilt FTS rows also index raw JSON tags instead of the space-joined form used at ingest (`src/pinky_daemon/kb_store.py:829`)
  - Removed the dead _content_preview body assignment and its comment; reindex now inserts ' '.join(json.loads(row['tags'])) to match ingest()/update_raw().
- **low** list(tag=...) filters tags in Python after SQL LIMIT/OFFSET - broken pagination and silently missing results (`src/pinky_daemon/presentation_store.py:424`)
  - Tag match moved into SQL via EXISTS json_each(presentations.tags) before LIMIT/OFFSET; Python post-filter removed.

- ResearchStore.list_topics post-LIMIT tag filter (mentioned inside the task_store tag-filter finding): research_store.py is not a file cited in my findings group, so editing it would violate the scope rule (owned by another branch). The path is also latent: routes/research.py never passes tag. Fixed only the cited task_store.py half.
- list_agents re-runs _compute_active_seconds once per agent (secondary mention in the pricing N+1 finding): The fix_hint targeted the pricing memoization, which I implemented. Restructuring list_agents' per-agent active-seconds query is a larger refactor with no hint coverage; deferring rather than risking behavior drift. The dominant per-row connection churn is gone.

## Testing
**stores**: All run from the worktree root. python3 -m pytest tests/test_analytics_store.py tests/test_kb_store.py tests/test_task_store.py tests/test_conversation_store.py tests/test_presentation_store.py -q -> 117 passed. Pre-fix verification (git stash push -- src/): all 12 new regression tests FAIL against unfixed code, pass after (12 failed -> 0). Broader sweeps: python3 -m pytest tests/test_api.py tests/test_agent_status.py tests/test_route_auth_coverage.py tests/test_auth.py -q -> 465 passed; tests/test_pinky_self.py tests/test_scheduler.py tests/test_hub.py tests/test_session_store.py -> 135 passed; tests/test_autonomy.py tests/test_agent_status.py -> 38 passed; tests/test_pinky_self_tools.py -> 180 passed; tests/test_shared_mcp_integration.py -> 10 passed. ruff check on all 10 changed files -> All checks passed. Zero failures anywhere.

### DIFF
--- a/src/pinky_daemon/analytics_store.py
+++ b/src/pinky_daemon/analytics_store.py
@@ -50,6 +50,7 @@ def _prev_range_bounds(range_name: str) -> tuple[str, str]:
 class AnalyticsStore:
     def __init__(self, db_path: str) -> None:
         self.db_path = db_path
+        self._pricing_cache: dict[tuple[str, str], list[dict]] | None = None
         self._init_db()
 
     @contextmanager
@@ -1053,6 +1054,8 @@ class AnalyticsStore:
                 WHERE started_at >= ? AND started_at <= ?
                   AND (tool_name = 'Bash' OR tool_name LIKE '%__Bash')
             """
+            if agent_name:
+                bash_query += " AND agent_name=?"
             bash_rows = conn.execute(bash_query, tool_params).fetchall()
 
         # Index tools and bash commands by (session_id, turn_seq)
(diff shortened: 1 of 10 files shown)


olegbrok (approved):

All 11 fixes verified as root-cause fixes with no schema migrations. Analytics crash: bash_query's binding-count mismatch fixed AND correctly scoped by agent_name, mirroring tool_query; #649 pricing parity preserved exactly (_lookup_pricing keeps identical alias order, range predicate, newest-first precedence; pricing.py untouched). FTS5 _fts5_phrase quote-doubling verified correct; recursive-CTE task delete + json_each tag pagination verified by execution; presentation MAX(version)+1 fixes a real UNIQUE collision after restore_version. Bonus: the ingest orphan-file cleanup addresses open side-finding #703.

Nits / follow-ups:
1. (minor) conversation_store.py:101 / kb_store.py:~615 — fallback catches ALL OperationalError; a locked-DB error silently becomes empty results with no log. Match on fts5/syntax messages or at least log.
2. (minor) kb_store.py:332-338 — dedup fix has no backfill: pre-existing rows keep old-format previews, so content dedup stays dead against historical sources until re-ingest; reindex() would be the natural hook.
3. (minor, cross-PR) confirmed pinky_memory/store.py:1018 has the same unescaped-MATCH pattern (degrades to LIKE, doesn't crash) — needs a follow-up issue since both sweeps avoid each other's files; #715 review notes it too.
4. (nit) _fts5_phrase duplicated in two stores (third copy wanted for pinky_memory) — shared util.
5. (nit) content-dedup goes live for the first time at two call sites — same-content-different-URL is now rejected; intended, but release-note it.

PR #718, opened 2026-06-10
fix(sweep): voice - TwiML escaping, WS teardown race, client leak

bradbrok:
## Summary

## Findings fixed
- **medium** welcomeGreeting XML attribute not quote-escaped - TwiML breakage / attribute injection (`src/pinky_daemon/voice_engine.py:262`)
  - build_outbound_twiml now embeds the greeting with saxutils.quoteattr(), so double quotes are escaped and attribute injection via the call goal is impossible.
- **medium** WS finally block unconditionally pops _active_ws and marks session completed (`src/pinky_daemon/voice_routes.py:861`)
  - conversationrelay_ws rejects a second WS for an already-connected session (close 4003), only pops _active_ws when the entry is its own ws, and a mark_completed flag skips the status=completed stamp (and finalize) on the callSid/accountSid setup-rejection paths.
- **medium** terminate_session reports {terminated: true} even when both hangup attempts failed (`src/pinky_daemon/voice_routes.py:930`)
  - Tracks ws_ended/rest_ended; if neither the WS end message nor the Twilio REST hangup succeeded it raises HTTPException 502 with the error and does not stamp the session completed.
- **low** finalize_call launched as fire-and-forget task; duplicate-artifact IntegrityError lost (`src/pinky_daemon/voice_routes.py:882`)
  - Finalize task is held in a module-level _finalize_tasks set with add_done_callback(discard) to prevent GC, and finalize_call (voice_engine.py) wraps save_artifact in try/except so a duplicate-call_sid IntegrityError is logged and the owner notification still goes out.
- **low** Dead code: HMAC approval-token helpers never called (`src/pinky_daemon/voice_store.py:63`)
  - Deleted make_approval_token/verify_approval_token and the now-unused hmac/hashlib/base64 imports; kept _APPROVAL_TOKEN_TTL which is still used for call_request.expires_at.
- **low** New unclosed AsyncAnthropic client created per conversation turn (`src/pinky_daemon/voice_engine.py:138`)
  - Added module-level _get_anthropic_client() cache keyed by api_key; haiku_respond and extract_outcome_with_opus now reuse one client (and its connection pool) per key instead of building one per turn.

## Testing
**voice**: python3 -m pytest tests/test_voice_engine.py tests/test_voice_routes.py -q -> 13 passed (includes new regression tests: TwiML quote-escaping + injection, duplicate-WS rejection, callSid-mismatch not marking completed, normal disconnect still marking completed, terminate 502 on dual failure, terminate success path, finalize survives save_artifact IntegrityError and still notifies, Anthropic client cache identity). python3 -m pytest tests/test_route_auth_coverage.py -q -> 7 passed. python3 -m pytest tests/test_api.py -q -> 351 passed. ruff check on all 5 changed files -> All checks passed. Import sanity: pinky_daemon.api/voice_* import ok.

### DIFF
--- a/src/pinky_daemon/voice_engine.py
+++ b/src/pinky_daemon/voice_engine.py
@@ -19,9 +19,25 @@ from typing import Any, AsyncIterator, Callable
 _HAIKU_MODEL = "claude-haiku-4-5-20251001"
 _OPUS_MODEL = "claude-opus-4-8"
 
+# Reuse one AsyncAnthropic client (and its connection pool) per API key so
+# each conversation turn doesn't pay a fresh TLS handshake.
+_anthropic_clients: dict[str, Any] = {}
+
 
 def _log(msg: str) -> None:
     print(msg, file=sys.stderr, flush=True)
+
+
+def _get_anthropic_client(api_key: str = "") -> Any:
+    """Get a cached AsyncAnthropic client for the given (or env) API key."""
+    import anthropic
+
+    key = api_key or os.environ.get("ANTHROPIC_API_KEY", "")
+    client = _anthropic_clients.get(key)
+    if client is None:
+        client = anthropic.AsyncAnthropic(api_key=key)
+        _anthropic_clients[key] = client
+    return client
 
 
 # ── Twilio helpers ───────────────────────────────────────────────────────────
@@ -132,10 +148,7 @@ async def haiku_respond(
     Yields text chunks as they arrive. Caller is responsible for
     sending them as CR text messages.
     """
-    import anthropic
-
-    key = api_key or os.environ.get("ANTHROPIC_API_KEY", "")
-    client = anthropic.AsyncAnthropic(api_key=key)
+    client = _get_anthropic_client(api_key)
 
     async with client.messages.stream(
         model=_HAIKU_MODEL,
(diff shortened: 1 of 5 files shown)


olegbrok (approved):

All six findings verified real with correct, minimally-scoped fixes. quoteattr is the right primitive — saxutils.escape leaves quotes, and req.goal reaches the greeting attribute without human review via AUTO_APPROVE_AGENTS, so the injection was genuinely reachable; regression test proves it. WS teardown: identity-checked pop + mark_completed + dup rejection (4003) is sound; no stranded call legs (Twilio status_callback independently stamps completion; CR WS close ends the call). Client leak: per-utterance AsyncAnthropic construction → daemon-lifetime cached singletons, correct on a single loop. Owner-approval gate untouched — the deleted HMAC token helpers were verified dead (zero callers). Zero overlap with #711 (broker-layer voice vs Twilio engine; no shared files).

Nits / follow-ups:
1. (minor) voice_routes.py:982-984 (pre-existing, untouched) — `<Dial>{brad_phone}</Dial>` interpolates unescaped element content; owner-controlled so not third-party-injectable, but a malformed value silently breaks transfer. Follow-up.
2. (minor) voice_engine.py:412-423 — finalize catches bare Exception around save_artifact (stated rationale is IntegrityError); disk-full/schema errors get swallowed into a log while reporting success. Narrow it.
3. (nit) terminate's ws_ended can false-positive (send_text success = buffered, not delivered); 502-on-dual-failure already strictly improves main; status callback reconciles.

PR #721, opened 2026-06-10
fix(sweep): cli - serve crash, connect clobbering settings, packaging fixes

bradbrok:
## Summary

## Findings fixed
- **medium** pinky serve always crashes: delegated MCP server mains re-parse pinky's own sys.argv (`src/pinky_cli/serve.py:22`)
  - Reset sys.argv to ["pinky-memory"]/["pinky-outreach"] before calling the delegated main(), mirroring the existing pattern in __main__.py's run branch.
- **low** pinky serve --server all (the default) only starts the memory server, never outreach (`src/pinky_cli/serve.py:19`)
  - server=="all" now spawns both servers as subprocesses via [sys.executable, "-m", ...] and waits on both, with terminate-on-KeyboardInterrupt; single-server modes remain in-process.
- **medium** pinky connect writes mcpServers into ~/.claude/settings.json, which Claude Code does not read for MCP servers (`src/pinky_cli/connect.py:14`)
  - connect now writes the mcpServers block to the project-level .mcp.json in cwd (the file Claude Code actually reads) and no longer touches ~/.claude/settings.json.
- **low** Re-running pinky connect without env vars clobbers previously working API keys with empty strings (`src/pinky_cli/connect.py:46`)
  - Env merge now keeps an existing non-empty credential when the env var is unset, omits empty env keys, and prints a warning for each missing credential.
- **low** pinky connect rewrites the user's entire global settings.json non-atomically and crashes on invalid JSON (`src/pinky_cli/connect.py:40`)
  - Invalid JSON in .mcp.json now aborts with a friendly message and exit(1) without writing; the config is written via a temp file plus os.replace for atomic replacement.
- **low** MCP server command hardcoded as 'python', which does not exist on stock macOS and ignores the active interpreter (`src/pinky_cli/connect.py:22`)
  - Both generated server entries now use sys.executable as the command, matching the pinky_daemon/api.py pattern.
- **low** pinky run silently ignores --config: daemon defaults to api mode and --mode is never passed (`src/pinky_cli/__main__.py:60`)
  - Added --mode {api,poll} (default api) to the run subparser and forwarded it in the rewritten sys.argv, making poll mode (and thus --config) reachable; --config help now says (poll mode).

- pinky init cannot find templates when installed as a wheel -- templates/ is not packaged: Correct fix requires out-of-scope edits: either [tool.hatch.build.targets.wheel] changes in pyproject.toml (force-include) or moving repo-root templates/CLAUDE.md.template and templates/pinky.yaml.example under src/pinky_cli/templates/. Only src/pinky_cli/init.py is in scope; editing init.py alone (e.g. importlib.resources lookup) cannot fix the bug because the template files are simply absent from the wheel, and duplicating the templates into the package would create two sources of truth. Defect confirmed (hatch wheel target lists only src/* packages, no force-include/shared-data); integrator should move templates into the package and resolve via importlib.resources.files("pinky_cli").

## Testing
**cli**: python3 -m pytest tests/test_cli.py -q -> 10 passed (new regression file covering serve argv reset, serve all subprocess spawn, connect .mcp.json write/credential preservation/warning/invalid-JSON abort/unrelated-key preservation, run --mode forwarding and default). ruff check src/pinky_cli/serve.py src/pinky_cli/connect.py src/pinky_cli/__main__.py tests/test_cli.py -> All checks passed (after one auto-fixed I001 import sort in the new test file). Live smoke tests: PYTHONPATH=src python3 -m pinky_cli serve --server memory </dev/null -> exit 0 (previously argparse exit 2); pinky_cli connect with/without env vars -> writes .mcp.json with sys.executable, preserves credentials, warns when unset; pinky_cli run --help shows --mode. grep of tests/ found no pre-existing tests referencing run_serve/run_connect/pinky_cli, so no existing assertions needed updating.

### DIFF
--- a/src/pinky_cli/__main__.py
+++ b/src/pinky_cli/__main__.py
@@ -33,9 +33,15 @@ def main() -> None:
     # run (daemon)
     run_parser = subparsers.add_parser("run", help="Run the Pinky daemon (headless Claude Code)")
     run_parser.add_argument(
+        "--mode",
+        default="api",
+        choices=["api", "poll"],
+        help="Run mode: api (HTTP server) or poll (message polling daemon)",
+    )
+    run_parser.add_argument(
         "--config",
         default="pinky.yaml",
-        help="Path to pinky.yaml config",
+        help="Path to pinky.yaml config (poll mode)",
     )
     run_parser.add_argument(
         "--working-dir",
@@ -57,7 +63,12 @@ def main() -> None:
     elif args.command == "run":
         from pinky_daemon.__main__ import main as daemon_main
         # Override sys.argv for the daemon's own arg parser
-        sys.argv = ["pinky-daemon", "--config", args.config, "--working-dir", args.working_dir]
+        sys.argv = [
+            "pinky-daemon",
+            "--mode", args.mode,
+            "--config", args.config,
+            "--working-dir", args.working_dir,
+        ]
         daemon_main()
     else:
         parser.print_help()
--- a/src/pinky_cli/connect.py
+++ b/src/pinky_cli/connect.py
@@ -4,29 +4,26 @@ from __future__ import annotations
 
 import json
 import os
+import sys
 from pathlib import Path
 
 
 def run_connect() -> None:
-    """Write MCP server config to Claude Code settings."""
-    home = Path.home()
-    settings_dir = home / ".claude"
-    settings_file = settings_dir / "settings.json"
-
-    # Build MCP server configs
+    """Write MCP server config to the project's .mcp.json (read by Claude Code)."""
     project_dir = Path.cwd().resolve()
+    mcp_file = project_dir / ".mcp.json"
     memory_db = project_dir / "data" / "memory.db"
 
     mcp_config = {
         "pinky-memory": {
-            "command": "python",
+            "command": sys.executable,
             "args": ["-m", "pinky_memory", "--db", str(memory_db)],
             "env": {
                 "OPENAI_API_KEY": os.environ.get("OPENAI_API_KEY", ""),
             },
         },
         "pinky-outreach": {
-            "command": "python",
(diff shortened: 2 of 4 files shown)


olegbrok (approved):

Serve crash is a true root-cause fix (delegated parse_args against pinky's own argv → argparse exit 2; the argv reset mirrors the run-branch pattern). Connect rewrite correctly retargets .mcp.json (the file Claude Code actually reads), with sound merge/atomic-write/abort semantics — verified missing-file, malformed-JSON, partial-keys, env-precedence, unrelated-server preservation. sys.executable fix is real (stock macOS has no `python`). Prod's repo-checkout deploy unaffected (launchd runs python -m pinky_daemon directly). The deferred templates-not-in-wheel defect is accurately diagnosed (hatch wheel target ships only src/*).

**CodeQL**: alert gate on ONE genuinely-new alert — #510, py/clear-text-logging-sensitive-data, connect.py:56 — but it's a false positive in substance: the warning prints the env-var NAME (never a value) and only fires when the value is empty/unset. Dismiss #510 as FP and the gate clears.

Nits:
1. (minor) connect.py:46-58 — merge preserves only the canonical env keys; user-added extras (e.g. DISCORD_BOT_TOKEN on pinky-outreach, which the server supports) are dropped — same clobbering class as the bug being fixed, narrower. Seed env from existing_env and overlay.
2. (minor) friendly-abort guards only JSONDecodeError; `[]`/`"x"`/null mcpServers still crash — add isinstance checks.
3. (minor) `serve all` children share one stdin so neither can serve MCP — fine as smoke-check (it was 100% broken before), but say so in help text.
4. (nit) .mcp.json now carries live tokens at default umask in a committable location — chmod 0600 + gitignore warning per #673 precedent.

PR #722, opened 2026-06-10
fix(sweep): fe-chat - XSS in markdown links, poll wiping scrollback, session rename

bradbrok:
## Summary

## Findings fixed
- **high** Stored XSS in chat via renderMarkdown links (javascript: URLs + attribute injection) (`frontend-svelte/src/lib/utils.js:91`)
  - escapeHtml now escapes double/single quotes (pure string replace) and markdown links only render as <a> for http(s)/mailto/relative URLs; unsafe schemes (javascript:, data:, vbscript:) are left as escaped plain text. Quote-escaping also closes the wiki-link data-attribute breakout.
- **medium** 3-second chat poll wipes older message pages loaded via infinite scroll (`frontend-svelte/src/pages/Chat.svelte:501`)
  - refreshChat now fetches limit=loadedPersistedCount (instead of always PAGE_SIZE) when the refresh targets the same conversation source and the user has paged in more than one page, so polling preserves the loaded history window (backend history endpoint has no limit cap; verified in conversation_store.py).
- **medium** Renaming the active session silently kills polling and SSE for the open chat (`frontend-svelte/src/pages/Chat.svelte:1156`)
  - finishRename now calls selectSession(`${agentName}-${newLabel}`, agentName) after a successful PATCH + refreshSessions when the renamed session was active, restarting chat polling, activity polling, and the EventSource against the new id.
- **medium** Agent name derived via split('-')[0] breaks for hyphenated agent names (`frontend-svelte/src/pages/Chat.svelte:1689`)
  - Added agentFromSessionId (longest-prefix match against agentsList) and labelFromSessionId helpers; replaced all split('-')[0] agent derivations (refreshSessions conv-sessions, refreshChat, startActivityPolling, search modal) and all split('-').slice(1).join('-') label derivations (tool-calls, effort, streaming status, SSE, session-meta, sendMessage, saveEffort, TmuxPaneModal) with them.
- **low** Sidebar label regex mangles session labels ending in 'main' (`frontend-svelte/src/components/ChatSidebar.svelte:64`)
  - Label now prefers s._streaming_label and otherwise strips only the agent-name prefix (dropped the /-?main$/ replace that turned 'domain' into 'do'); '<agent>-main' still displays as 'main' via the existing || 'main' fallback, so rename PATCH targets the real label.
- **low** 'connected' flag is dead state - daemon-unreachable condition never shown to user (`frontend-svelte/src/pages/Chat.svelte:365`)
  - Added a 'daemon unreachable - showing cached data' banner at the top of the chat area bound to !connected, and refreshSessions' catch now logs the error via console.error (matching existing frontend convention).
- **low** persistedToolCalls not reset on session switch and its error path ignores the refresh sequence guard (`frontend-svelte/src/pages/Chat.svelte:497`)
  - selectSession now clears persistedToolCalls alongside liveToolCalls, and the tool-call fetch catch only wipes persistedToolCalls when requestSeq === chatRefreshSeq && sessionId === activeSession, matching the other guarded catches.

## Testing
**fe-chat**: cd frontend-svelte && npm install --no-audit --no-fund (ok, 130 packages); npm run build (success, 206 modules transformed, only pre-existing warnings about chunk size/dynamic import); node smoke test of the XSS fix importing renderMarkdown/escapeHtml directly: [click](javascript:alert`1`) no longer renders an anchor, the "onmouseover= attribute-breakout payload stays inside href as &quot;, wiki-link attributes are quote-escaped, and normal https/relative links still render. No frontend test suite exists; no Python code was touched so pytest was not run.

### DIFF
--- a/frontend-svelte/src/components/ChatSidebar.svelte
+++ b/frontend-svelte/src/components/ChatSidebar.svelte
@@ -61,7 +61,7 @@
             </div>
             {#each aSessions as s}
                 {@const isMain = (s.session_type || '') === 'main'}
-                {@const label = s.id.replace(new RegExp(`^${agent.name}-`), '').replace(/-?main$/, '') || 'main'}
+                {@const label = s._streaming_label || s.id.replace(new RegExp(`^${agent.name}-`), '') || 'main'}
                 {@const isRenaming = renamingSession && renamingSession.agentName === agent.name && renamingSession.label === label}
                 <div
                     class="session-item"
--- a/frontend-svelte/src/lib/utils.js
+++ b/frontend-svelte/src/lib/utils.js
@@ -62,9 +62,19 @@ export function contextClass(pct) {
 }
 
 export function escapeHtml(text) {
-    const div = document.createElement('div');
-    div.textContent = text;
-    return div.innerHTML;
+    return String(text ?? '')
+        .replace(/&/g, '&amp;')
+        .replace(/</g, '&lt;')
+        .replace(/>/g, '&gt;')
+        .replace(/"/g, '&quot;')
+        .replace(/'/g, '&#39;');
+}
+
+// Allow http(s)/mailto and scheme-less relative URLs; blocks javascript:,
+// data:, vbscript: etc. from becoming clickable hrefs.
+function isSafeLinkUrl(url) {
+    if (/^(?:https?|mailto):/i.test(url)) return true;
+    return !url.split(/[/?#]/, 1)[0].includes(':');
 }
 
 export function renderMarkdown(text) {
@@ -88,7 +98,8 @@ export function renderMarkdown(text) {
 
     const applyInline = (value) => {
         let html = escapeHtml(value);
-        html = html.replace(/\[([^\]]+)\]\(([^)\s]+)\)/g, '<a href="$2" target="_blank" rel="noreferrer">$1</a>');
+        html = html.replace(/\[([^\]]+)\]\(([^)\s]+)\)/g, (match, label, url) =>
+            isSafeLinkUrl(url) ? `<a href="${url}" target="_blank" rel="noreferrer">${label}</a>` : match);
         // [[Wiki Link]] → clickable wiki cross-link
         html = html.replace(/\[\[([^\]]+)\]\]/g, (_, title) => {
             const slug = title.toLowerCase().replace(/\s+/g, '-');
--- a/frontend-svelte/src/locales/en.json
+++ b/frontend-svelte/src/locales/en.json
@@ -313,7 +313,8 @@
     "session": "Session",
     "show_agents": "Show Agents",
     "upload_file": "Upload file",
-    "search_placeholder": "Search chats..."
+    "search_placeholder": "Search chats...",
+    "offline_banner": "daemon unreachable - showing cached data"
   },
   "fleet": {
     "search_placeholder": "Search agents...",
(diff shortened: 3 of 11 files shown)


olegbrok (changes requested):

The XSS fix is genuinely complete for the chat/research/KB markdown sink — escapeHtml now covers quotes (closing the href/data-attr breakouts), isSafeLinkUrl allowlists schemes case-insensitively, and I checked entity-smuggling: escaping runs before link matching and no dangerous scheme survives — the prompt-injection-to-XSS chain is closed. Rename fix is correct and complementary to #710's backend half (disjoint files, either merge order). Hyphenated-agent longest-prefix matching verified across all 12 call sites.

1. **(major) Chat.svelte:419-423 + :545 — `historyLimit()` is an unbounded fetch ratchet.** `loadedPersistedCount` is set from the merged list AFTER up to 20 session-event messages are mixed in, so on any chat with ≥100 stored messages + ≥1 session event (i.e. every active agent's main chat, no paging needed), each 3s poll grows the next fetch by ~E until it's refetching the ENTIRE conversation history every 3 seconds (the history endpoint has no limit cap — verified api.py:4452) plus a full client re-sort/re-render. Open tab degrades badly over hours. Fix: track store-loaded count separately from the merged display list (same uninflated count should drive loadOlderMessages' offset, which is also skewed today), and/or cap historyLimit (~1000).
2. (minor) window is anchored at newest, so new arrivals while reading scrollback shift absolute scrollTop slightly — far better than the old wipe; offset-anchored fetch would be exact.
3. (minor) Memories.svelte:115 — residual same-class {@html} sink: statsHtml interpolates the by_type key unescaped; safe today only because ReflectionType is an enum — give it escapeHtml(). (Audited all other {@html} sinks: covered or static.)
4. (nit) offline-banner string hardcoded English vs $_() elsewhere.

Fix the ratchet and this is a clean approve.

bradbrok:
Major + minors 3/4 addressed in a1f65e6:

- Major (both halves): refreshChat captures the store-loaded count BEFORE session events are merged, so loadedPersistedCount is a fixed point under polling - limit only grows when the user explicitly pages older (loadOlderMessages adds exactly the store rows it fetched, fixing the offset skew too), hard-capped at 1000.
- Minor 3: Memories.svelte statsHtml by_type key wrapped in escapeHtml() - #723 owns the file but its hunks are elsewhere; merge-tree verified clean both orders.
- Minor 4: offline banner moved to $_('chat.offline_banner'), key added to all seven locale files.

Skipped minor 2 (offset-anchored fetch) as the follow-up the review allows. npm run build green; manual walk-through of the poll/merge logic in the commit body explains why the ratchet is closed.

olegbrok (approved):
## Barsik re-review — approve ✅

All findings fixed at a1f65e63:
1. Ratchet broken at the source: `nextStoreCount` captured BEFORE the session-events merge and that's what feeds `loadedPersistedCount` (line 478/546), plus a `HISTORY_LIMIT_MAX = 1000` cap in `historyLimit()` as defense-in-depth. `loadOlderMessages` offset now uses the uninflated count (no skipped rows when paging), and both cache paths (save/restore) store/derive the clean count. All four entry points consistent.
2. Memories.svelte by_type key now escapeHtml()'d.
3. Offline banner i18n'd with real translations in all 7 locales.

utils.js untouched — the prior XSS verification stands. One non-blocking design note: paging beyond 1000 then polling collapses scrollback to the cap; state stays self-consistent and bounding is the explicit intent — fine.

PR #723, opened 2026-06-10
fix(sweep): fe-rest - dead filters/toggles, wizard state leaks, provider key redaction

bradbrok:
## Summary

## Findings fixed
- **medium** Skill Catalog never loads on page mount (`frontend-svelte/src/pages/Settings.svelte:703`)
  - Added refreshSkills() to the onMount loader block so the catalog populates on page open.
- **medium** Wizard does not reset wizTelegramRef/wizDiscordRef/wizSlackRef (`frontend-svelte/src/pages/Agents.svelte:999`)
  - openWizard() now resets all three wiz*Ref vars alongside the wiz*Token resets, so a canceled run's token selection is never applied to the next agent.
- **medium** Memory salience filter and sort dropdown silently do nothing (`frontend-svelte/src/pages/Memories.svelte:130`)
  - Renamed query params to salience_min/salience_max/sort_by to match the backend route signature in routes/memory.py.
- **low** Agent wizard never fetches global bot tokens (`frontend-svelte/src/pages/Agents.svelte:999`)
  - openWizard() now fetches /bot-tokens into globalBotTokens (mirroring the /providers fetch) so the Outreach step dropdowns work on a fresh install.
- **low** Trigger 'last fired' timestamp double-converted to ms (`frontend-svelte/src/pages/Agents.svelte:2118`)
  - Dropped the *1000 at the timeAgo callsite; timeAgo already converts numeric unix seconds to ms.
- **low** Release-channel 'beta' button always fails (`frontend-svelte/src/pages/Settings.svelte:810`)
  - Removed the dead stable/beta toggle and switchChannel()/channelSwitching; channel now renders as static text (serverInfo.channel || 'stable').
- **low** Group-create and retire modals close before the API call (`frontend-svelte/src/pages/Agents.svelte:469`)
  - submitGroup() and confirmRetire() now await the API in try/catch, toast the error and keep the modal open on failure, closing only after success.
- **low** openDetail sub-loaders lack the request-sequence guard (`frontend-svelte/src/pages/Agents.svelte:578`)
  - All 14 sub-loaders capture openDetailSeq before their await and bail before assigning state if superseded; closeDetail() bumps the seq to invalidate in-flight loaders.
- **low** Sidebar 'Active agents' list never renders (`frontend-svelte/src/components/Layout.svelte:49`)
  - Unwraps agentsResp.agents from the {agents,count,main_agent} envelope; status dot now keys off agent.streaming (agent.status never existed).
- **low** Activity feed collapses 'Load more' results on every 10s auto-refresh (`frontend-svelte/src/pages/Dashboard.svelte:103`)
  - refresh() sizes the activity refetch from activityOffset (raw fetched count) instead of the filtered activityEvents.length, so the expanded list no longer shrinks each tick.
- **n/a** group_note: provider_key_set migration (write-only provider keys) (`frontend-svelte/src/pages/Agents.svelte, frontend-svelte/src/pages/Settings.svelte`)
  - Key fields prefill empty and track (provider_key_set ?? provider_key) booleans, show an 'API key (saved) - leave blank to keep it' hint, and PUT bodies include provider_key only when the user typed a new value; tolerant of both old and new backend response shapes.

- 'Active only' toggle is non-functional - inactive memories can never be shown: The only correct fix is backend-side (make routes/memory.py active param tri-state, MemoryQueryFilters.active Optional in pinky_memory/types.py, gate the WHERE clause in store.py) - all out-of-scope files owned by parallel branches. A frontend-only change (sending active=false when the toggle is off) would show ONLY inactive memories instead of all memories, which is wrong semantics, so I did not force it.

## Testing
**fe-rest**: cd frontend-svelte && npm install --no-audit --no-fund: succeeded (130 packages). npm run build: succeeded twice (1.7s / 1.6s, 206 modules transformed); only pre-existing warnings (chunk >500kB, ineffective dynamic import of i18n.js). No frontend test suite exists; no Python files were touched so pytest was not run. Verified all added lines are ASCII-only (git diff grep for non-ASCII: clean).

### DIFF
--- a/frontend-svelte/src/components/Layout.svelte
+++ b/frontend-svelte/src/components/Layout.svelte
@@ -46,7 +46,7 @@
             ]);
             statusText = `v${root.version}`;
             authenticated = !!auth.authenticated;
-            if (Array.isArray(agentsResp)) agents = agentsResp;
+            agents = Array.isArray(agentsResp?.agents) ? agentsResp.agents : [];
 
             // First-run: redirect to onboarding if no agents and not yet completed
             if (authenticated) {
@@ -140,7 +140,7 @@
                             class="sidebar-nav-item sidebar-agent"
                             on:click={closeSidebar}
                         >
-                            <span class="agent-dot" class:alive={agent.status === 'running'} class:idle={agent.status !== 'running'}></span>
+                            <span class="agent-dot" class:alive={agent.streaming} class:idle={!agent.streaming}></span>
                             <span>{agent.name}</span>
                         </a>
                     {/each}
--- a/frontend-svelte/src/pages/Agents.svelte
+++ b/frontend-svelte/src/pages/Agents.svelte
@@ -98,6 +98,7 @@
     // Model provider state
     let providerUrl = '';
     let providerKey = '';
+    let providerKeySet = false; // backend reports provider_key_set instead of the plaintext key
     let providerModel = '';
     let providerPreset = 'anthropic'; // 'anthropic' | 'ollama' | 'zai' | 'openrouter' | 'deepseek' | 'codex_cli' | 'custom'
     let providerRef = '';  // ID of a global provider (empty = use agent-specific config)
@@ -466,8 +467,13 @@
         const members = groupMembers.split(',').map(m => m.trim()).filter(Boolean);
         if (!groupName.trim()) { toast('Group name required', 'error'); return; }
         if (!members.length) { toast('Add at least one member', 'error'); return; }
+        try {
+            await api('POST', '/groups', { name: groupName, members });
+        } catch (e) {
+            toast('Failed to create group: ' + (e?.message || e), 'error');
+            return;
+        }
         groupModalOpen = false;
-        await api('POST', '/groups', { name: groupName, members });
         toast(`Group "${groupName}" created`);
         refreshAgents();
     }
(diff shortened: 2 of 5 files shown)


olegbrok (approved):
## Barsik review — approve (merge-with-nits) — deploy BEFORE or WITH #712

10 of 11 fixes verified exactly against the main backend (param names, response envelopes, timestamp units, seq-guards). This PR IS the frontend follow-up #712's review asked for: all three provider_key reads migrated to provider_key_set, shape-tolerant in both merge orders (`?? ` fallback). Preferred deploy order: **#723 before #712** — if #712 lands first, old frontend shows 'no key' badges until this arrives.

1. **(major, cross-PR — needs #712-side fix, flagged there too) 'Reset to Anthropic default' no longer clears a stale backend key in ANY merge order.** saveProvider now omits falsy provider_key, and post-#712 even explicit '' is ignored — so an agent reset to Anthropic keeps exporting a leftover foreign key as ANTHROPIC_API_KEY and is permanently mis-detected as 'custom' on reopen. Cheap partial mitigation HERE: when the user picked anthropic/global-ref, send provider_key:'' explicitly (clears on pre-#712 main, harmless no-op after). Real fix: backend null clear sentinel in #712.
2. (minor) Layout.svelte:49 — 'Active agents' uses GET /agents default enabled_only=false, so disabled agents show with an idle dot; use ?enabled_only=true.
3. (minor) 'API key (saved)' hint requires providerUrl truthy — key-only configs (legal) show an empty password field with no hint.
4. (nit) two hint strings hardcoded English in otherwise-i18n files.
5. (nit) inverted-vs-direct seq-guard style in catch blocks — pick one.

Memories filters / trigger timestamp / wizard leaks / beta-button removal / dashboard offset / skill-catalog onMount all verified correct.

PR #724, opened 2026-06-10
fix(sweep): runners-comms - claude CLI --prompt flag, timeout reaping, resume semantics

bradbrok:
## Summary

## Findings fixed
- **high** ClaudeRunner passes prompt via nonexistent --prompt flag; every CLI invocation fails (`src/pinky_daemon/claude_runner.py:236`)
  - Pass the prompt as the final positional argument instead of cmd.extend(["--prompt", prompt]); updated tests/test_daemon.py which asserted the buggy flag and now pins cmd[-1] == prompt and absence of --prompt.
- **medium** ClaudeRunner timeout path never kills the subprocess (`src/pinky_daemon/claude_runner.py:173`)
  - On asyncio.TimeoutError, process.kill() (catching ProcessLookupError) then await process.wait() to reap the child before returning the timeout RunResult; regression test asserts the spawned proc has returncode set.
- **medium** codex exec stderr pipe never drained during the turn (`src/pinky_daemon/codex_session.py:595`)
  - Start stderr_task = asyncio.create_task(proc.stderr.read()) right after spawn and await it after proc.wait() (mirrors codex_app_server._drain_stderr); finally-block cancels it on error paths. Regression test wedges pre-fix (child writes 256KiB to stderr before any stdout).
- **medium** idle_sleep() runs _exec_codex directly, racing the message worker (`src/pinky_daemon/codex_session.py:1363`)
  - Added self._exec_lock = asyncio.Lock(); both the worker's exec call and idle_sleep's save-turn exec acquire it, so two execs can never run concurrently against the same codex thread. Regression test asserts max concurrent execs == 1.
- **low** App-server mode emits assistant text twice to the UI stream (`src/pinky_daemon/codex_session.py:1060`)
  - _handle_event's agent_message branch skips the full-text assistant_delta emit when self._use_app_server (deltas already streamed it); text_parts append stays unconditional. Tests cover both app-server dedupe and legacy path still emitting.
- **low** User-message conversation-store append swallows errors with no log (`src/pinky_daemon/codex_session.py:282`)
  - Replaced except Exception: pass with a _log of the append failure, matching the assistant-side handler and StreamingSession.
- **low** Unguarded proc.kill() in exec timeout handler can raise and discard the turn result (`src/pinky_daemon/codex_session.py:685`)
  - Wrapped proc.kill()/await proc.wait() in the TimeoutError handler in try/except Exception: pass, mirroring the sibling except-Exception branch.
- **low** turn_usage per-turn tracking is collected but never used (`src/pinky_daemon/sdk_runner.py:203`)
  - Deleted the dead turn_usage declaration and the per-AssistantMessage append block; nothing read it (verified by repo-wide grep).
- **medium** Per-chat session ids are not UUIDs and resume semantics break chat isolation - poll-mode pipeline cannot work (`src/pinky_daemon/message_handler.py:116`)
  - _get_session_id now derives deterministic per-chat UUIDs (uuid5); ClaudeRunner uses --resume <id> for continuation instead of --session-id + --continue, passes the prompt positionally (the --prompt flag does not exist), and falls back to creating the session with --session-id when --resume reports 'No conversation found' (first message of a chat / fresh disk).
- **medium** get_thread hides group and broadcast messages from their recipients (`src/pinky_daemon/agent_comms.py:527`)
  - Session-filtered thread query now also accepts inbox delivery (OR i.session_id IS NOT NULL on the already-session-scoped LEFT JOIN), so group/broadcast recipients see the thread; unrelated sessions still see nothing.
- **medium** Unvalidated parent_message_id enables cyclic threads that hang the recursive CTE (`src/pinky_daemon/agent_comms.py:517`)
  - _send raises ValueError when parent_message_id does not reference an existing message (rejects self/forward/guessed ids, so cycles cannot be created), and both recursive CTEs use UNION instead of UNION ALL so cyclic legacy data terminates instead of generating rows forever.
- **medium** send_file builds the destination path from unsanitized to_session - path traversal and cwd-relative writes (`src/pinky_daemon/agent_comms.py:340`)
  - to_session is validated (rejects empty, '.', '..', and anything containing path separators), the destination is realpath-contained as a direct child of the transfers root, and the root is anchored next to the comms DB (Path(db_path).resolve().parent/'transfers') instead of cwd-relative 'data' - same location for the daemon, but cwd-independent.
- **low** get_all_messages duplicates each multi-recipient message and breaks pagination (`src/pinky_daemon/agent_comms.py:559`)
  - Audit query now GROUP BY m.id with COALESCE(MAX(i.read), 0), so each message appears once and LIMIT/OFFSET pagination is consistent with the total count.
- **low** mark_read with an explicit empty list marks the entire inbox as read (`src/pinky_daemon/agent_comms.py:411`)
  - mark_read distinguishes None (mark all) from [] (mark nothing, return 0), matching the docstring contract; also avoids the IN () SQL syntax error an empty list would otherwise hit.

- get_thread via run_in_executor (sub-item of the cyclic-CTE finding's fix_hint): Moving the synchronous sqlite3 call off the event loop requires editing api.py, which is owned by parallel branches. The hang itself is fixed at the source: parents are validated at send time and the CTE uses UNION, so the query is always finite.

## Testing
**runners**: python3 -m pytest tests/test_codex_session.py tests/test_daemon.py tests/test_claude_runner_binary.py -q -> 113 passed (re-run after ASCII cleanup: 110 passed across the first two files). python3 -m pytest tests/test_api.py tests/test_tmux_session.py -q -> 606 passed. python3 -m pytest tests/test_sdk_runner.py tests/test_effort_verification.py -q -> 56 passed. python3 -m pytest tests/test_tmux_dream_runner.py -q -> 12 passed. ruff check on all 5 changed files -> All checks passed. New regression tests: test_run_timeout_kills_subprocess (test_daemon.py), TestExecSerialization.test_idle_sleep_save_exec_serializes_with_worker, TestExecStderrDrain.test_large_stderr_does_not_wedge_exec, TestAssistantDeltaDedupe (2 tests) (test_codex_session.py); test_build_command_basic updated to pin the positional-prompt contract.
**comms**: python3 -m pytest tests/test_agent_comms.py tests/test_daemon.py tests/test_claude_runner_binary.py -q -> 99 passed. python3 -m pytest tests/test_api.py tests/test_pinky_self_tools.py tests/test_route_auth_coverage.py -q -> 538 passed (6m10s). Re-run after ASCII header cleanup: python3 -m pytest tests/test_agent_comms.py tests/test_daemon.py -q -> 96 passed. ruff check on all 5 changed files -> All checks passed. New regression tests: 8 in tests/test_agent_comms.py (group/broadcast thread visibility, unrelated-session denial, unknown-parent rejection, cyclic-parent termination, traversal rejection, transfers-root containment, audit dedup + pagination, empty-list mark_read) and 4 in tests/test_daemon.py (resume-with-session flag shape, create-on-resume-miss fallback, UUID/determinism of session ids); 3 existing tests updated (--prompt removal, uuid5 session id expectations). Existing claude CLI semantics verified empirically: --session-id creates (errors 'already in use' on reuse), --resume <uuid> resumes (errors 'No conversation found' on stderr when missing).

### DIFF
--- a/src/pinky_daemon/agent_comms.py
+++ b/src/pinky_daemon/agent_comms.py
@@ -65,6 +65,9 @@ class AgentComms:
 
     def __init__(self, db_path: str = "data/agent_comms.db") -> None:
         self._db_path = db_path
+        # Anchor file transfers next to the DB so the destination does not
+        # depend on the caller's cwd at send time.
+        self._transfers_root = Path(db_path).resolve().parent / "transfers"
         Path(db_path).parent.mkdir(parents=True, exist_ok=True)
         self._conn = sqlite3.connect(db_path, check_same_thread=False)
         self._conn.row_factory = sqlite3.Row
@@ -272,6 +275,15 @@ class AgentComms:
         ttl_seconds: int | None = None,
     ) -> AgentMessage:
         """Internal: store message and deliver to recipients' inboxes."""
+        if parent_message_id is not None:
+            parent = self._conn.execute(
+                "SELECT 1 FROM messages WHERE id = ?", (parent_message_id,)
+            ).fetchone()
+            if parent is None:
+                raise ValueError(
+                    f"parent_message_id {parent_message_id} does not reference an existing message"
+                )
+
         ts = time.time()
         meta_json = json.dumps(metadata or {})
 
@@ -336,8 +348,18 @@ class AgentComms:
         if not os.path.isfile(file_path):
             raise FileNotFoundError(f"Source file does not exist: {file_path}")
 
+        if (
+            not to_session
+            or to_session in (".", "..")
+            or os.path.basename(to_session) != to_session
+        ):
+            raise ValueError(f"Invalid file transfer recipient: {to_session!r}")
+
         filename = os.path.basename(file_path)
-        dest_dir = os.path.join("data", "transfers", to_session)
+        dest_root = os.path.realpath(self._transfers_root)
+        dest_dir = os.path.realpath(os.path.join(dest_root, to_session))
+        if os.path.dirname(dest_dir) != dest_root:
+            raise ValueError(f"Invalid file transfer recipient: {to_session!r}")
         os.makedirs(dest_dir, exist_ok=True)
 
         dest_path = os.path.join(dest_dir, filename)
@@ -408,7 +430,9 @@ class AgentComms:
         Returns:
             Number of messages marked.
         """
-        if message_ids:
+        if message_ids is not None:
+            if not message_ids:
+                return 0
             placeholders = ",".join("?" * len(message_ids))
             cursor = self._conn.execute(
                 f"""UPDATE inbox SET read = 1
@@ -510,21 +534,25 @@ class AgentComms:
                 break
             root_id = row["parent_message_id"]
 
-        # Recursive CTE to get all descendants at any depth.
+        # Recursive CTE to get all descendants at any depth. UNION (not
+        # UNION ALL) so cyclic parent chains in legacy data terminate.
         # Left-join inbox to get the real read state for this session
         # (falls back to False when the session has no inbox entry).
+        # A session is authorized to see a message if it sent it, is the
+        # direct recipient, or had it delivered to its inbox (group and
+        # broadcast messages store the group name or "*" in to_session).
         if session_id:
             query = """
                 WITH RECURSIVE thread AS (
                     SELECT m.* FROM messages m WHERE m.id = ?
-                    UNION ALL
+                    UNION
                     SELECT m.* FROM messages m
                     JOIN thread t ON m.parent_message_id = t.id
                 )
                 SELECT t.*, COALESCE(i.read, 0) as read
                 FROM thread t
                 LEFT JOIN inbox i ON i.message_id = t.id AND i.session_id = ?
-                WHERE t.from_session = ? OR t.to_session = ?
+                WHERE t.from_session = ? OR t.to_session = ? OR i.session_id IS NOT NULL
                 ORDER BY t.timestamp ASC
             """
             params: list = [root_id, session_id, session_id, session_id]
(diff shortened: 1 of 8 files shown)


olegbrok (approved):

**Dream-risk for tomorrow's 7:10am tmux dream verification: none** — tmux_dream_runner/dream_runner untouched; tmux dream path imports only the unchanged RunResult dataclass; sdk_runner change is verified-dead turn_usage removal (grepped both trees); codex changes affect Codex agents only.

CLI flag fix verified against the actual claude CLI: `--prompt` doesn't exist (only --prompt-suggestions), prompt is positional, --session-id creates, --resume resumes — the legacy poll-mode pipeline was 100% dead before this. Timeout reaping now actually reaps (real /bin/sleep test); agent_comms fixes (path traversal containment, CTE cycle termination via UNION, empty-list mark_read marking EVERYTHING read) all verified with good edge tests; get_thread's OR-clause is safe (LEFT JOIN already scoped to the requesting session — no authz widening). codex _exec_lock covers both transport paths, no nested acquisition.

Nits / follow-ups:
1. (minor) claude_runner.py:193-199 — kills only the direct child; CLI-spawned MCP-server children get orphaned (the #663 orphan class). Pass start_new_session=True + os.killpg; TERM→KILL escalation would be nicer than straight SIGKILL.
2. (minor) daemon.py:179 (untouched) — poll-mode autonomy session_sender still passes pinky-<hex12> ids; --resume fails with an error that does NOT match the 'No conversation found' fallback string, so autonomy wakes in poll mode stay broken. Only the platform-message path got the uuid5 migration. Not a regression (was 100% dead), but the PR claim is half-true — follow-up.
3. (minor) the 'No conversation found' stderr string-match is brittle across CLI upgrades — comment it with the pinned CLI version.
4. (nit) positional prompt appended without `--` separator; a prompt starting with '-' parses as a flag.

PR #782, opened 2026-06-15
[codex] add one-click containerize UI

olegbrok:
## Summary
- Replace the manual Runtime-tab isolation select/image field with a one-click containerize panel.
- Add runtime preflight handling for `GET /system/container-runtime`, async status polling for `GET /agents/{name}/container-status`, and actions for `POST /agents/{name}/containerize` / `decontainerize`.
- Surface LOCAL/CONTAINERIZED/operation badges on agent cards and in the detail Runtime tab.

## Notes
- This is the frontend half of the one-click containerize work and expects the backend API contract agreed with Barsik.
- If the backend endpoints are not available yet, the UI fails closed: container controls show unavailable instead of breaking the Agents page.
- The old manual isolation PUT path is removed from the Runtime tab; transport remains visible because it is still useful operational context.

## Validation
- `npm ci`
- `npm run build`

Build warnings observed are existing Vite warnings: large chunk size and `src/lib/i18n.js` ineffective dynamic import.

### DIFF
--- a/frontend-svelte/src/pages/Agents.svelte
+++ b/frontend-svelte/src/pages/Agents.svelte
@@ -403,7 +403,7 @@
 
     async function handleContainerActionError(agentName, e, retry) {
         const msg = e?.message || String(e);
-        if (msg.startsWith('409:') && confirm(`${msg}\n\nForce the transition and stop active sessions?`)) {
+        if (msg.startsWith('409:') && msg.toLowerCase().includes('busy') && confirm(`${msg}\n\nForce the transition and stop active sessions?`)) {
             await retry(true);
             return;
         }
--- a/frontend-svelte/src/pages/Agents.svelte
+++ b/frontend-svelte/src/pages/Agents.svelte
@@ -107,6 +107,13 @@
     let thinkingEffortDirty = false;
     let detailIsolationMode = 'local';   // 'local' | 'container' — detail Runtime tab editor state
     let detailContainerImage = '';       // OCI image ref shown/edited in the detail Runtime tab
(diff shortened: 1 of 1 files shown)


olegbrok (commented):
**Review (Barsik) — LGTM** ✅ (mutual review per Brad)

Solid frontend. Verified the critical paths:
- **409 force-confirm works**: `handleContainerActionError` keys on `e.message.startsWith('409:')`, and `api.js:37` throws `` `${resp.status}: ${detail}` `` — so the busy → confirm → force-retry flow fires correctly.
- **No timer leak**: `onDestroy` calls `stopAllContainerStatusTimers()`; `pollContainerStatus` clears its interval once status leaves the active set.
- **Fails closed**: `loadContainerRuntime` catch → `unavailable:true`, controls show unavailable instead of breaking the page when the backend endpoints aren't present.
- **Contract match**: preflight fields, the `containerStatusActive` set, the 202-as-initial-payload handling, and `containerPayload` (`image` only when overridden) all line up with the backend (#783).
- Actionable `containerizeDisabledReason`, badges on cards + Runtime tab, old `saveAgentIsolation` path removed.

Non-blocking nits (defer freely):
1. `handleContainerActionError` force-confirms on **any** 409, including the backend's "operation already in progress" 409 where `force` won't help (button is disabled during in-flight, so it's a corner case). Could gate the confirm on the message mentioning "busy".
2. Card badges only reflect live op state while the detail modal is open (only detail polls); cards otherwise show static LOCAL/CONTAINERIZED from `isolation_mode`. Fine as a design choice.

Backend (#783) addressed your two P1s in 48081d8. Once CI's green on both + your re-review, we line up the merges.

🤖 Reviewed by Barsik

PR #806, opened 2026-06-21
fix(auth): allow per-agent Claude auth mode override

olegbrok:
## Summary
- add per-agent `PINKY_CLAUDE_AUTH_MODE_<AGENT>` override with fallback to global `PINKY_CLAUDE_AUTH_MODE`
- thread agent-aware auth-mode resolution through tmux spawn and both credential seed gates
- cover default/global/per-agent/invalid override behavior and guardrail seed suppression via tests

## Verification
- `uv run ruff check src/pinky_daemon/tmux_session.py tests/test_tmux_container_isolation.py`
- `git diff --check`
- `uv run pytest tests/test_tmux_container_isolation.py -q` (70 passed)
- `uv run pytest tests/test_tmux_session.py -q` (282 passed)

### DIFF
--- a/src/pinky_daemon/tmux_session.py
+++ b/src/pinky_daemon/tmux_session.py
@@ -161,23 +161,47 @@ _CLAUDE_AUTH_MODES = {
 }
 
 
-def _claude_auth_mode() -> str:
-    """Fleet auth mode for Claude Code tmux sessions.
+def _claude_auth_mode_env_for_agent(agent_name: str | None) -> str | None:
+    if not agent_name:
+        return None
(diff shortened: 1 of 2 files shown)


olegbrok (commented):
Reviewed — I independently built the identical patch before our messages crossed, so this is a high-confidence review. Precedence (per-agent → global → default), agent-name sanitization, and self.agent_name threading through both seed gates + spawn telemetry are all correct. Your invalid-per-agent → global fallback (via _resolve_claude_auth_mode) is cleaner than my version's. LGTM. Discarded my duplicate; merging on CI green.

Issue #249, opened 2026-04-17
[P-00] Federation v0.2 Umbrella / Tracking Issue


olegbrok:
**Planning ID:** `P-00`  
**Milestone:** `ga`  
**Labels:** `federation`, `repo:pinky`, `tracking`  

Parent tracking issue for Federation v0.2. Single status surface for milestone progress, risks, and readiness.

## Repos
- PinkyBot (client): [bradbrok/PinkyBot](https://github.com/bradbrok/PinkyBot)
- Federation relay (server): [olegbrok/pinkybot-federation-relay](https://github.com/olegbrok/pinkybot-federation-relay)

## Hard product decisions (Apr 17, 2026)
- **No recovery, no escrow.** Tenant signing key is client-owned; loss = rebuild tenant from scratch. Enforced in UI (P-07) and docs (P-13).
- **Tiered metadata retention.** Transport TTL is tier-aware: 7d default / 30d paid. Implemented at R-03. Pricing copy parked in R-07.
- **Recipient-consented excerpts.** Any abuse-report excerpt is decrypted client-side and uploaded as an explicit, consented artifact (P-11 + R-08). Relay never sees plaintext as a side effect.

### `crypto-spike`

**PinkyBot:**
- [ ] #250 — Implement `sealed_box_v1` keyring and envelope crypto primitives
- [ ] #252 — Implement TOFU pinning and fingerprint mismatch state machine

**Relay:**
- [ ] olegbrok/pinkybot-federation-relay#1 — Implement public-key directory and key-bundle publish/fetch APIs

### `alpha`

**PinkyBot:**
- [ ] #251 — Add local federation state store and secure tenant/device key persistence
- [ ] #253 — Build two-instance federation integration harness
- [ ] #254 — Implement federation client runtime with encrypted outbox/inbox delivery loop
- [ ] #255 — Add first federation MCP tools for send, discover, and key lookup
- [ ] #256 — Build Fleet UI for create-tenant and join-tenant flows

**Relay:**
- [ ] olegbrok/pinkybot-federation-relay#2 — Build tenant bootstrap and instance enrollment APIs
- [ ] olegbrok/pinkybot-federation-relay#3 — Implement authenticated WebSocket transport, store-and-forward queueing, and tier-aware envelope TTL

### `private-beta`

**PinkyBot:**
- [ ] #257 — Build Fleet invite issuance, join completion, and membership surfaces
[...]

PR #820, opened 2026-06-24
feat(slack): purchase-approval buttons + interactivity wiring (#249 phase 1)

olegbrok:
## #249 phase 1 — Slack purchase-approval buttons + interactivity wiring

Adds interactive **Approve / Reject** buttons to the pos-spec-purchasing approval
card, with a deterministic, **fail-closed approver gate in the daemon**. Replaces
(augments) the "type `approve` in the channel" flow.

Slack approval card

> Faithful render of the actual `build_proposal_blocks` output (can't screenshot
> live Slack until the Pi deploy + Interactivity toggle). Emoji-free per house rule.

### Why this shape
The verified, unspoofable approver identity (`payload['user']['id']`, signed by
Slack) exists **only** at the daemon's Socket Mode handler. Once a click is handed
to an LLM, that identity is just text a model could be injected into mis-stating.
So the **authoritative money-gate lives in the daemon**, before anything reaches
the agent. The daemon has no MCP client (it can't call `approve_purchase`
directly), so an authorized click is routed to the owning agent to execute — the
LLM is an executor, **never the security boundary**.

### What's in here
**Block Kit send path** (Slack `blocks`): `pinky-messaging send()` gains a
`blocks` param (JSON string), threaded through `/broker/send` → `_broker_send`
→ `_send_text_message` → `SlackAdapter.send_message` → `chat.postMessage`.
Malformed JSON degrades to text-only; `blocks` folded into the send-dedupe key.

**Interactivity** (`block_actions`): `BrokerSlackPoller` now handles `interactive`
envelopes (were ACK'd + dropped). `_handle_interactive`:
- The **GATE** — verified clicker id vs `registry.is_purchase_approver`.
  Fail-closed: blank id / unconfigured allowlist / any error → **deny**.
  Unauthorized clicks get an ephemeral refusal and **never reach the agent**.
- `pending_id` / `clicker_id` validated as clean opaque tokens at the boundary
  (fail-closed) → injection-safe agent instruction.
- Authorized → inject a synthetic instruction to the owning agent to call
  `approve_purchase`/`reject_purchase` with the **verified** approver, then update
  the card via `response_url` (buttons removed; approver name emoji-stripped).
- `SlackAdapter.respond_via_url()` for the card update (capability URL, no scope).

**Allowlist** (authoritative gate): `AgentRegistry.get/set/is_purchase_approver`
(`purchase_approver_slack_ids` system setting), fail-closed.

### Enforcement layers
1. **Daemon gate (authoritative)** — verified Slack id vs allowlist, fail-closed.
2. **MCP allowlist (secondary)** — `POS_PURCHASING_APPROVERS` in pos-spec-purchasing
   (separate PR), only as trustworthy as the approver string passed; a guard, not
   the gate.

### Phase-1 scope / phase-2
Phase 1 `approve_purchase` returns only an Amazon **cart link — no money moves**.
The legacy text-`approve` path is intentionally **not** gated by the verified-id
check; in phase 1 the worst case is a spurious cart link. **Phase 2**
(pay-on-approval) closes that: the daemon records the verified click and the
payment tool refuses to pay without a matching record. **Do not enable real
payment before that exists.** See `specs/slack-purchase-approval.md`.

### Tests
`tests/test_slack_purchase_approval.py` — gate authz / fail-closed (no registry,
registry error, blank/unknown approver) / route-to-agent / agent-offline /
malformed-token drop / emoji strip / Cyrillic preserved; registry allowlist;
adapter blocks + `response_url`. All green; ruff clean. Existing Slack/broker/
messaging suites unaffected.

### Review
Adversarial multi-lens review run before this PR (bypass/spoofing, fail-closed,
correctness/regression, contract/Block-Kit). Fail-closed + correctness lenses
clean; bypass + contract findings fixed (boundary token validation, emoji strip,
Slack-limit clips on the MCP side).

### Deploy prerequisites (Pi / chekov) — coordinated separately
1. Enable the Slack app **Interactivity & Shortcuts** toggle (Socket Mode already
   on → no Request URL, no reinstall).
2. Seed the allowlist (fail-closed): `registry.set_purchase_approvers(["U7W8RJGP5"])`.
3. Restart is on the **Pi** (chekov's poller) — heads-up to Brad first.


olegbrok:
**Security hardening since review** (commit afd1cb0, pos-spec-purchasing#1 @ 2450edf)

Addressed the blocking review finding: the daemon gate is now genuinely enforced, not just claimed. A purchase state transition requires a **daemon-signed HMAC token** the daemon mints only after the verified-clicker gate passes; the MCP verifies it (`POS_PURCHASING_APPROVAL_SECRET`, exact {decision, pending_id, approver, expiry} tuple, not expired, constant-time) before any `approve_purchase`/`reject_purchase` state change. An LLM cannot forge it, so **no prompt path — including the legacy text-`approve` path — can self-approve**, and reject is gated too. Fail-closed when the shared secret is unset on either side.

Cross-repo HMAC contract locked by a golden-vector test in both repos. Added `test_server_approval.py` (tool-boundary enforcement via `call_tool`). Spec updated: the token is now the authoritative layer. CI re-running on afd1cb0.

PR #280, opened 2026-04-19
feat(presence): server-side last_seen stamping for non-CC agents

olegbrok:
## Summary

Fixes #279 — non-CC agents (Codex, GLM) appeared offline because they never call the `send_heartbeat` MCP tool. Presence is now stamped server-side on real activity (turn completion, inbound message delivery) so the registry reflects actual liveness regardless of whether the agent manually heartbeats.

## Changes

- **`agent_registry.py`** — new `agents.last_seen_at REAL` column via `_ensure_columns` migration; `AgentRegistry.stamp_last_seen(name, ts=None)` helper.
- **`api.py`** — `_agent_presence` now returns `max(agent.last_seen_at, heartbeat.timestamp)`; registry plumbed into session init.
- **`broker.py`** — stamp after successful `inject_agent_message` send and inbound `_route_streaming` delivery.
- **`streaming_session.py`** — stamp once after turn lifecycle (both success and error paths).
- **`codex_session.py`** — same pattern on `turn.completed` / `turn.failed` for Codex/GLM.
- Stamp failures are logged but never block delivery or turn completion — presence is telemetry, not critical path.

## Test plan

- [x] `pytest tests/test_agent_registry.py tests/test_broker.py` — 47 passed
- [x] Full suite: 1475 passed, 1 skipped (pinky_federation skipped — PyNaCl missing, pre-existing)
- [x] Ruff clean on touched files
- [ ] Manual verify: Codex/GLM agent shows online in dashboard after a turn without calling send_heartbeat

## Notes

- `register()` INSERT untouched — `DEFAULT 0` on the new column handles inserts automatically.
- Turn stamp collapsed into a single post-branch call (vs duplicating at `turn_completed` and `turn_error`) since both prove the pipe is alive.

Closes #279

🤖 Opened by Barsik

Issue #279, opened 2026-04-19
Server-side presence tracking: stamp last_seen agent-agnostic


olegbrok:
## Problem

`last_seen` on agent presence is populated from the `send_heartbeat` MCP tool (api.py:5175-5188). Only Claude Code agents call this tool. Non-CC agents (Codex-based murzik, GLM-based agents) have a live streaming session but their `last_seen` shows stale (e.g. murzik last_seen=2026-04-14 despite active session).

The streaming-session-is-active signal IS correct — `broker._streaming` has the entry and `is_connected=true`. It's the heartbeat-derived `last_seen` that's wrong for non-CC agents.

## Fix (option 1, greenlit)

Server-side, agent-agnostic presence stamping. The daemon stamps `last_seen` on:

1. Successful `broker.inject_agent_message` delivery (agent alive enough to receive)
2. Streaming turn completion (agent alive enough to reason)

Keep the existing `send_heartbeat` tool working for CC agents — no regression.

## Files

- `src/pinky_daemon/api.py` — `_agent_presence` at L5175, `send_agent_message_direct` at L6872
- `src/pinky_daemon/broker.py` / streaming session — turn completion hook
- Agent registry — add helper `stamp_last_seen(agent_name)` updating `agents.last_seen`

## Acceptance

- murzik (Codex) + GLM agents show fresh `last_seen` while streaming
- No regression for Barsik/Pushok/etc. heartbeat-reporting agents
- Tests cover both code paths (inject delivery, turn completion)

🤖 Opened by Barsik

PR #831, opened 2026-06-28
fix(broker): agent-reply auto-route as true fallback with dedup (#280)

olegbrok:
## What

Make the #279 agent-to-agent reply **auto-route a true fallback** instead of an unconditional second delivery path, eliminating duplicate inbox entries when an agent replies explicitly.

## Why

#279 delivers a completed agent-injected turn's final text to the requester's inbox. But agents (Murzik especially) also reply explicitly via `send_to_agent` → `comms.send`, which writes straight to the inbox. A turn that did **both** delivered the reply **twice**.

The dup hadn't surfaced in the wild yet only because no agent-injected turn had completed since #279 deployed (verified against the live `conversations_agent_comms.db`: the only `auto_routed` row in 3 days is a pushok self-test; every Murzik verdict came via explicit `send_to_agent` and predates the deploy). The next review verdict would have doubled.

## How

- `inject_agent_message` stamps the turn boundary: `(responder, requester) -> start_ts`.
- `route_agent_reply` pops the marker and asks `comms.has_message_since(responder, requester, since, exclude_auto_routed=True)` — i.e. *did the responder already send an explicit reply to the requester during this turn?* If yes → **suppress** (the explicit copy is already in the inbox). If no → **deliver** (genuine fallback, e.g. an interrupted turn that produced text but never sent it).
- **Backend-agnostic** — no reliance on `tool_uses` (codex/tmux turns omit them), which is why detection is done against the comms DB, not the turn payload.
- **Never content-matches** — so it can't drop a genuinely new reply.
- **Fails safe** — a missing marker (lost across a restart) or a probe error falls through to delivery. A stray dup is acceptable; a dropped review verdict is not.
- Orphan markers swept by TTL (1h) on insert → bounded memory.
- Kill-switch `PINKY_AGENT_REPLY_DEDUP` (default on); off restores the #279 unconditional delivery. Counted via the existing broker `deduped` stat.

## Design note for review

This uses a **timestamp window**, not content equality. Rationale: the auto-route is a *safety net for agents that produce a reply but don't send it*. If an agent demonstrably knows how to send (it sent explicitly to the requester this turn), the net is off for that turn — it should send any additional content explicitly too. Trade-off: if an agent sends an *unrelated* explicit message mid-turn (e.g. a clarifying question) and then relies on the auto-route for its actual answer, the answer is suppressed (recoverable — it's in the transcript and the requester notices the gap). Considered acceptable vs. content-matching, which would silently miss real dups whenever a codex turn's final text differs from the sent content.

## Tests

`tests/test_broker.py` — suppress-on-explicit-send, deliver-on-no-send, **deliver-on-missing-marker (safety)**, killswitch-off-always-delivers, inject-marks-turn-start.
`tests/test_agent_comms.py` — `has_message_since` incl. exclude-auto-routed, window boundary, legacy/empty-metadata-counts.

`pytest tests/test_broker.py tests/test_agent_comms.py` → 126 passed. `ruff` clean.

## Deploy

**Held** — not for immediate deploy. Touches the live agent-reply path; wants Murzik review + a supervised daemon bounce (recent fan-out instability on the Mini). Inert behind the existing #279 routing being on; the dedup defaults on but only changes behavior on the explicit-send-this-turn path.

🤖 Opened by Barsik

### DIFF
--- a/src/pinky_daemon/agent_comms.py
+++ b/src/pinky_daemon/agent_comms.py
@@ -456,6 +456,38 @@ class AgentComms:
         ).fetchone()
         return row[0]
 
+    def has_message_since(
+        self,
+        from_session: str,
+        to_session: str,
+        since_ts: float,
+        *,
+        exclude_auto_routed: bool = True,
+    ) -> bool:
+        """Whether a direct message from ``from_session`` to ``to_session`` was
+        stored at or after ``since_ts``.
+
+        Backs the #280 agent-reply dedup: the broker asks "did the responder
+        already send the requester an explicit reply during this turn?" before
+        auto-routing the same turn's text. ``exclude_auto_routed`` skips the
+        broker's own auto-routed copies (metadata ``{"auto_routed": true}``) so
+        only deliberate ``send_to_agent`` sends count. Existence probe — returns
+        as soon as one match is found.
+        """
+        query = (
+            "SELECT 1 FROM messages "
+            "WHERE from_session = ? AND to_session = ? AND timestamp >= ?"
+        )
+        if exclude_auto_routed:
+            # COALESCE so rows with empty/legacy metadata (NULL json_extract)
+            # are treated as explicit (not auto-routed) and DO count.
+            query += " AND COALESCE(json_extract(metadata, '$.auto_routed'), 0) = 0"
+        query += " LIMIT 1"
+        row = self._conn.execute(
+            query, (from_session, to_session, since_ts)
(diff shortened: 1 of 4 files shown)


olegbrok:
Closing as superseded: #1075 removed implicit turn-final delivery entirely (the auto-route fallback this PR de-duplicated no longer exists; explicit outreach tools are the only delivery path), so the double-delivery case cannot occur on main.

Issue #506, opened 2026-05-16
Expand _protected_api_prefixes to currently-unlisted routes (defense-in-depth)


olegbrok:
## Context

PR #504 landed a **scoped** default-deny in `auth_middleware` (`src/pinky_daemon/api.py`): only paths matching `_protected_api_prefixes` are flat-401'd at the bottom of the middleware. Everything else falls through to `call_next`.

The scoped form was the right call for #504's scope (the global default-deny in v2 blocked `/calendar/google/callback` before its state-nonce validation could run — see Murzik's PR #504 round-2 review). But it leaves a gap: routes that aren't in `_protected_api_prefixes` AND aren't in `_public_*paths` are reachable without auth. Most of those routes today are NOT auth-by-state — they're protected-by-omission.

This is the same bug class as the original #497 — a route is unauth-reachable because nobody updated a list. The list is just shorter now.

## Current state

`_protected_api_prefixes` (as of 426a516):

```
"/agents", "/sessions", "/tasks", "/projects", "/research",
"/presentations", "/presentation-templates", "/skills", "/outreach",
"/system", "/settings", "/scheduler", "/activity", "/audit",
"/conversations", "/heartbeats", "/groups", "/hooks", "/autonomy",
"/broker", "/kb", "/migration", "/auth/password", "/api/voice"
```

## Currently unlisted routes (audit, partial — needs verification)

Each of these is registered via `@router.get/post/...` somewhere in `src/pinky_daemon/routes/*` or `api.py` but doesn't match any prefix in `_protected_api_prefixes`. Most should be added; a few should be carved out as public.

**Likely should be added to `_protected_api_prefixes`:**
- `/analytics/*` — analytics endpoints
- `/apps/*` — app management (deploy, share, password, upload, etc.)
- `/bot-tokens/*` — bot token management
- `/calendar/*` (except OAuth callback below) — calendar config/status
- `/comms/*` — comms inboxes/messages
- `/milestones/*` — project milestones
- `/sprints/*` — project sprints
- `/models/*` — LLM model config
- `/providers/*` — provider config
- `/plugins/*` — plugin management
- `/render/*` — PDF render
- `/soul-templates/*` — soul template management
- `/triggers` — trigger management
- `/user-profiles/*` — user profile management
- `/parse`, `/parse_dir`, `/preview`, `/apply` — migration tool routes

**Should remain public (auth-by-state or auth-by-token in path):**
[...]

PR #421, opened 2026-05-10
feat(ferry): outbound mesh_remote_send tool (#418) — stacked on #414

olegbrok:
Closes #418. **Stacked PR** — branched off \`feat/ferry-host-pinky\` so it can share \`ferry/types.py\`. Once #414 merges, rebase target → \`beta\` is a no-op.

## Summary

Outbound counterpart to #414's \`host_pinky.py\` inbound. Build canonical ferry envelopes per PROTOCOL.md v0.1 and publish them to the mesh transport, exposed to agents as the \`mesh_remote_send\` MCP tool on pinky-self.

End-to-end wire was proven earlier today (Barsik smoke 2026-05-10 12:14 PDT, sub-2s, FERRY_MESSAGES seq #3 confirmed by @Pulse). This PR replaces the raw \`nats\` CLI scaffolding with a properly-permissioned tool path.

## Why shellout (not nats-py)

\`host_pinky.py\`'s docstring is explicit that the Python module is the *adapter* and \"does no transport (NATS / leaf-link plumbing is owned by ferry-core).\" To keep that boundary clean on outbound, we shell out to the \`nats\` CLI rather than introduce a new in-process transport. One-shot publishes match the CLI shape exactly.

If we later need a long-lived in-process publisher, \`MeshSender.send\` is the swap point — callers don't change.

## What lands (5 files, +999 LOC)

| File | Adds |
|------|------|
| \`src/pinky_daemon/ferry/outbound.py\` | \`parse_address\`, \`derive_subject\`, \`allowlist_matches\`, \`build_envelope\`, \`envelope_to_wire\`, \`MeshSender\` (new) |
| \`src/pinky_daemon/agent_registry.py\` | \`mesh_outbound_allowlist\` column + accessors mirroring peer_fleet_acl pattern |
| \`src/pinky_daemon/api.py\` | \`MeshSendRequest\` + \`MeshAllowlist*Request\` models, \`GET/POST/PUT/DELETE /agents/{name}/mesh/allowlist\`, \`GET /mesh/diagnostics\`, \`POST /agents/{name}/mesh/send\` |
| \`src/pinky_self/server.py\` | \`mesh_remote_send(target, body, kind, …)\` MCP tool next to \`send_to_agent\` |
| \`tests/test_ferry_outbound.py\` | 44 tests (new) |

## Cred handling

Read from env at \`MeshSender\` construction time:
[...]

PR #891, opened 2026-07-18
Authenticate dream-runner skill proposals (#421)

olegbrok:
## Problem

`DreamRunner._extract_proposed_skills` POSTs to `/skills/from-md` with no auth headers, but `/skills` is in the deny-by-default auth-gated prefix list (#506). Every dream-proposed skill has been silently dying with 401 fleet-wide (observed as onesie@TOD's skill-create failures, task #420/#421).

## Fix

- Sign the POST with `build_internal_auth_headers`, resolving the secret via `resolve_request_signing_secret` (existing #623 plumbing): per-agent signing key when a resolver is injected (isolated agents), global `PINKY_SESSION_SECRET` fallback otherwise. Resolver errors degrade to the fallback — never raise into the dream pipeline.
- `create_api` injects `AgentRegistry.get_signing_key` as the resolver.

## Verification

- New regression test builds the real `create_api` app and hits the live auth gate: bare POST → 401; signed POST → 200 + skill created, parameterized for both the non-isolated global-fallback path and the isolated per-agent-key path.
- `tests/test_dream_runner.py` 15/15; auth + route-coverage suites 89/89 at this commit (Barsik, clean worktree). Murzik: 104 relevant tests pass, SHA verified.
- Built by Kuzya, reviewed by Murzik, verified by Barsik (task #421, spun out of #420 investigation).

Backend-only change — no UI surface; evidence is the test output above.

🤖 Opened by Barsik

### DIFF
--- a/src/pinky_daemon/api.py
+++ b/src/pinky_daemon/api.py
@@ -2491,6 +2491,7 @@ def create_api(
         ),
         owner_provider=lambda: agents.get_owner_profile(),
         setting_provider=lambda key: agents.get_setting(key, ""),
+        signing_key_provider=agents.get_signing_key,
     )
     app.state.dream_runner = dream_runner
     app.state.agent_history_resolver = _resolve_agent_history
--- a/src/pinky_daemon/dream_runner.py
+++ b/src/pinky_daemon/dream_runner.py
@@ -15,6 +15,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import os
 import re
 import sqlite3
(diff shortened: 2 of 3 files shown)


bradbrok (approved):
Approved at exact reviewed SHA 57facd2cc3d41f5c30b6b732a9debf56ef6cee32. Independent verification: 104 relevant tests passed (dream runner, auth, and route-auth coverage), with ruff and diff checks clean.

PR #918, opened 2026-07-27
Deprecate agent inbox delivery

bradbrok:
## Summary

Deprecates the agent inbox delivery system in favor of live injection only.

- Agent messages are retained in `messages` for audit/thread history, with legacy `inbox` rows kept read-only (`read=1`) only for recipient authorization/bookkeeping.
- Startup retires any legacy unread rows; inbox/read/summary surfaces are inert and `check_inbox` returns a deprecation notice.
- Removes wake-time inbox replay and unread reminder generation.
- Makes auto-routed agent replies a one-way live injection (`route_reply=False`) so the existing inbox-only return leg is not lost and cannot ping-pong.
- Reports a failed per-call transport handoff as not delivered; `confirmed` remains observability only.

## Design choice

This is a clean behavioral deprecation rather than a feature flag. The SQLite schema is retained surgically because group/broadcast thread authorization and audit history use recipient mappings; those rows can never become unread delivery state.

## Validation

- `ruff check .`
- `python3 -m compileall -q src/pinky_daemon src/pinky_self`
- affected regression: 709 passed
- CPython 3.12.13 full repository suite with declared `.[dev]` dependencies: 4,425 passed / 3 skipped
- `git diff --check`

The production daemon already owns shared-MCP port 8890 on this host; one test that deliberately enables shared MCP emits an expected thread warning, without a test failure.

## Deployment

No merge, release, deploy, daemon restart, or live write performed. Barsik owns merge, release cut, and `update_and_restart`.

### DIFF
--- a/src/pinky_daemon/agent_comms.py
+++ b/src/pinky_daemon/agent_comms.py
@@ -6,7 +6,9 @@ Enables sessions (agents) to send messages to each other:
 - Broadcast: to all active agents
 
 Messages are stored in SQLite for durability and searchability.
-Each session has an inbox that accumulates messages until read.
+Delivery is handled separately by the daemon's live-injection path. The legacy
+inbox schema remains as read-only recipient bookkeeping for audit/thread
+authorization, but it is never exposed as a delivery inbox.
 """
 
 from __future__ import annotations
@@ -148,6 +150,11 @@ class AgentComms:
             self._conn.execute(
                 "CREATE INDEX IF NOT EXISTS idx_inbox_expires ON inbox(expires_at)"
             )
+            retired = self._conn.execute(
+                "UPDATE inbox SET read = 1 WHERE read = 0"
+            ).rowcount
+            if retired:
+                _log(f"agent_comms: retired {retired} legacy unread inbox row(s)")
             self._conn.execute("COMMIT")
         except Exception:
             self._conn.execute("ROLLBACK")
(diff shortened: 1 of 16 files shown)


bradbrok (commented):
PINNED INDEPENDENT GAUNTLET — PASS

Frozen coordinates:
- head: 1e6c7c63d763da9019f3ef98fac5cc633beb691b
- tree: 600c34bacfd33210f23f01c116c202d242a354a5
- sole parent / base / merge-base: 5848de0659fb1c5460e8a7637a4979eab38eed6d
- branch: agent/deprecate-agent-inbox
- divergence: 0 behind / 1 ahead

I validated this exact commit from a wholly fresh network clone, detached at the pinned SHA with a clean tracked tree. The 16-file (+330/-571) delta behaviorally deprecates durable agent-inbox delivery: new recipient bookkeeping is always read=1, legacy unread rows retire at startup, inbox/read/summary and check_inbox compatibility surfaces are inert, wake replay and unread nags are absent, failed live handoffs are not reported queued, and the auto-routed return leg uses route_reply=False for a one-way live injection while retaining audit/thread authorization bookkeeping.

Independent results:
- CPython 3.11.15 affected regression: 709 passed
- CPython 3.12.13 full repository suite: 4,425 passed / 3 skipped
- CPython 3.13.14 affected regression: 709 passed
- outside-repository adversarial harness on 3.11 / 3.12 / 3.13: 12/12 each
- Ruff; compileall over src + tests; git diff --check; git fsck --full --strict: PASS
- sdist + wheel build; fresh CPython 3.12 wheel install/import/core behavior: PASS
- GitHub frozen-head checks: Python 3.11/3.12/3.13, Ruff, frontend+i18n, CodeQL Python/JS, Docker amd64 smoke and arm64 build all SUCCESS (release-only multi-arch push intentionally skipped)

The full local suite emitted only the known non-failing thread warning when the one shared-MCP test attempted port 8890 already owned by the production daemon.

No blocking findings.

This PASS is non-transferable: any head, tree, parent/base, merge-base, or branch-ref drift requires a wholly fresh detached exact-SHA gauntlet and a new pinned artifact. No merge, release, deploy, restart, staging, runner, or live-system write was performed.

PR #925, opened 2026-07-28
Add in-place wake schedule updates

bradbrok:
## Summary

- add `AgentRegistry.update_schedule()` for ID-stable partial updates, with empty-update refusal and the same cron validation used by schedule creation
- expose an ownership-checked `PATCH /agents/{agent_name}/schedules/{schedule_id}` route and a gated `update_wake_schedule` MCP tool
- reject creation when an enabled schedule already has the same `(agent_name, name)`, while allowing reuse after disablement or one-shot completion
- cover registry, API, MCP wrapper, and shared-MCP gate behavior in pytest

## Why

Schedule prompt or timing corrections currently require create-new + delete-old, which renumbers durable schedule references. Repeated `set_wake_schedule` calls can also leave two enabled same-name rows that both fire.

## Impact

Agents can now edit a schedule in place without changing its ID. PATCH requests cannot edit another agent's schedule, and duplicate enabled-name creation fails closed with guidance to use `update_wake_schedule`.

Fixes #924.

## Validation

- focused schedule/API/MCP/shared-gate matrix: `40 passed`
- full mocked pinky-self tools plus shared gate checks: `206 passed`
- broader scheduler/API/pinky-self/shared/auth-route matrix: `799 passed`, with one production-environment-only dream transport failure; the exact selector passes on both this branch and untouched `origin/main` under CI-default SDK/shared-MCP flags
- `uv run ruff check .`
- `python3 -m compileall -q src/pinky_daemon src/pinky_self`
- `git diff --check`
- added-line width audit: all lines at or below 100 columns

🤖 Opened by Kuzya

### DIFF
--- a/src/pinky_daemon/agent_registry.py
+++ b/src/pinky_daemon/agent_registry.py
@@ -33,6 +33,7 @@ import time
 from dataclasses import dataclass, field
 from pathlib import Path
 
+from pinky_daemon.cron_utils import _field_matches
 from pinky_daemon.effort import is_ultracode
 
 # Agent names appear in filesystem paths (data/agents/{name}/, hook scripts
@@ -257,8 +258,6 @@ def _cron_next_run(cron: str, timezone: str = "UTC") -> float | None:
         import datetime as dt
         import zoneinfo
 
-        from pinky_daemon.scheduler import _field_matches
-
         parts = cron.strip().split()
         if len(parts) != 5:
             return None
@@ -302,8 +301,6 @@ def _cron_next_run(cron: str, timezone: str = "UTC") -> float | None:
 
 def _validate_schedule_cron(cron: str) -> None:
     """Reject cron expressions that the scheduler cannot match."""
-    from pinky_daemon.scheduler import _field_matches
-
     parts = cron.strip().split()
     if len(parts) != 5:
         raise ValueError("Cron expression must contain exactly five fields")
--- a/src/pinky_daemon/agent_registry.py
+++ b/src/pinky_daemon/agent_registry.py
@@ -2913,6 +2913,30 @@ except Exception:
 
     # ── Schedules ───────────────────────────────────────────
 
+    def _ensure_schedule_name_available(
+        self,
+        agent_name: str,
+        name: str,
+        *,
+        exclude_schedule_id: int | None = None,
+    ) -> None:
+        """Refuse an enabled agent/name collision, optionally excluding one row."""
+        sql = """SELECT id FROM agent_schedules
+                 WHERE agent_name=? AND name=? AND enabled=1"""
+        params: list = [agent_name, name]
+        if exclude_schedule_id is not None:
+            sql += " AND id<>?"
+            params.append(exclude_schedule_id)
+        sql += " ORDER BY id ASC LIMIT 1"
+        existing = self._db.execute(sql, params).fetchone()
+        if existing:
+            raise ScheduleNameConflictError(
+                f"Enabled schedule {name!r} already exists for agent {agent_name!r} "
+                f"as ID {existing[0]}; choose a distinct name to create another "
+                f"schedule, or use update_wake_schedule with ID {existing[0]} "
+                "to edit the existing one"
+            )
+
     def add_schedule(
         self, agent_name: str, cron: str, *,
         name: str = "", prompt: str = "", timezone: str = "America/Los_Angeles",
@@ -2921,17 +2945,7 @@ except Exception:
     ) -> AgentSchedule:
         """Add a cron-based wake schedule for an agent."""
         _validate_schedule_cron(cron)
-        existing = self._db.execute(
-            """SELECT id FROM agent_schedules
-               WHERE agent_name=? AND name=? AND enabled=1
-               ORDER BY id ASC LIMIT 1""",
-            (agent_name, name),
-        ).fetchone()
-        if existing:
-            raise ScheduleNameConflictError(
-                f"Enabled schedule {name!r} already exists for agent {agent_name!r} "
-                f"as ID {existing[0]}; use update_wake_schedule to edit it"
-            )
+        self._ensure_schedule_name_available(agent_name, name)
 
         now = time.time()
         cursor = self._db.execute(
@@ -3010,6 +3024,19 @@ except Exception:
             if column in updates:
                 updates[column] = int(updates[column])
 
+        current = self._db.execute(
+            "SELECT agent_name, enabled FROM agent_schedules WHERE id=?",
+            (schedule_id,),
+        ).fetchone()
+        if current is None:
+            return None
+        if name is not None and bool(current[1]):
+            self._ensure_schedule_name_available(
+                current[0],
+                name,
+                exclude_schedule_id=schedule_id,
+            )
+
         set_clause = ", ".join(f"{field}=?" for field in updates)
         cursor = self._db.execute(
             f"UPDATE agent_schedules SET {set_clause} WHERE id=?",
@@ -3034,6 +3061,18 @@ except Exception:
 
     def toggle_schedule(self, schedule_id: int, enabled: bool) -> bool:
         """Enable/disable a schedule."""
+        current = self._db.execute(
+            "SELECT agent_name, name FROM agent_schedules WHERE id=?",
+            (schedule_id,),
+        ).fetchone()
+        if current is None:
+            return False
+        if enabled:
+            self._ensure_schedule_name_available(
+                current[0],
+                current[1],
+                exclude_schedule_id=schedule_id,
+            )
         cursor = self._db.execute(
             "UPDATE agent_schedules SET enabled=? WHERE id=?",
             (int(enabled), schedule_id),
--- a/src/pinky_daemon/agent_registry.py
+++ b/src/pinky_daemon/agent_registry.py
@@ -300,6 +300,23 @@ def _cron_next_run(cron: str, timezone: str = "UTC") -> float | None:
         return None
 
 
+def _validate_schedule_cron(cron: str) -> None:
+    """Reject cron expressions that the scheduler cannot match."""
+    from pinky_daemon.scheduler import _field_matches
+
+    parts = cron.strip().split()
+    if len(parts) != 5:
+        raise ValueError("Cron expression must contain exactly five fields")
+
+    limits = ((0, 59), (0, 23), (1, 31), (1, 12), (0, 6))
+    try:
+        for part, (lo, hi) in zip(parts, limits):
+            if not any(_field_matches(part, value, lo, hi) for value in range(lo, hi + 1)):
+                raise ValueError
+    except (TypeError, ValueError) as exc:
+        raise ValueError(f"Invalid cron expression: {cron!r}") from exc
+
+
 # Injected into the system prompt when an agent's effective effort is the
 # ``ultracode`` tier (#151). Replicates Claude Code's native ultracode
 # semantics — "xhigh effort plus standing dynamic-workflow orchestration" —
@@ -634,6 +651,10 @@ class AgentSchedule:
             "target_channel": self.target_channel,
             "one_shot": self.one_shot,
         }
+
+
+class ScheduleNameConflictError(ValueError):
+    """An enabled schedule already uses the requested agent/name pair."""
 
 
 @dataclass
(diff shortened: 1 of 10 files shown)


bradbrok (commented):
# PinkyBot PR #925 — independent exact-pin review

Verdict: **FAIL — fix required**

Reviewed only the frozen target:

- PR: `bradbrok/PinkyBot#925`
- head: `3c52b9d4b187d06fb280ac6c353ebfb92f6affab`
- tree: `7f0efd8c011658de0cba8cc6970e15ef4a057898`
- sole parent/base/merge-base: `4c178dce2ce921415a68e29c2f36e30efb2d9a77`
- base tree: `7b5aecd6b32f2f207af9d26d3ab9b23956548144`
- patch: one commit, eight files, `+505/-14`

At review bind the PR was OPEN+DRAFT and MERGEABLE, with the remote head and
base still at the frozen SHAs. Nine checks were successful, the intentionally
non-PR multi-arch push job was skipped, and Python 3.12 CI was still running.
Any head or tree movement invalidates this verdict.

## Findings

1. **Blocking — the new update path bypasses the enabled-name conflict
   invariant.** `AgentRegistry.update_schedule()` accepts `name`, constructs
   its fixed-key update set, and writes it without checking whether another
   enabled row already owns the resulting `(agent_name, name)` pair
   (`agent_registry.py:2982-3018`). Starting with enabled schedules named
   `same` and `other`, `update_schedule(other.id, name="same")` succeeds and
   leaves both IDs enabled under `same`. The PATCH route and
   `update_wake_schedule` expose that operation directly. The scheduler
   iterates enabled rows independently, so both rows can fire. This defeats
   the duplicate-fire safety property and is especially contradictory because
   the create conflict tells callers to use `update_wake_schedule`.

   The same invariant can be bypassed after the required disabled-name reuse:
   disable schedule A, create enabled schedule B under A's name, then toggle A
   back on. Both A and B become enabled. Enforce enabled `(agent_name, name)`
   uniqueness on every mutation that can establish it: create, an enabled
   row's rename, and enable/re-enable. A rename check must exclude the row
   being edited. Return `ScheduleNameConflictError` consistently (including a
   409 from PATCH rather than the current generic 400 path), and add registry
   and API regression tests for rename and re-enable.

2. **Contract guidance required — unnamed creates now share one unique
   name.** Raw API creates that omit `name` all use `""`; MCP
   `set_wake_schedule` calls that omit it all use `"self_scheduled"`. A second
   enabled unnamed create therefore returns 409. That is a defensible
   fail-closed consequence of the requested same-name guard, not silent data
   loss, but it is a material tool-contract change. The MCP description still
   presents `name` as optional without saying enabled names are unique, and
   the conflict message only recommends editing the existing row. Document
   that concurrent distinct schedules require distinct names and make the
   conflict guidance distinguish “choose a distinct name to create another”
   from “use update_wake_schedule to edit this one.”

## Passed areas

1. **Partial-update behavior: PASS.** IDs and omitted values are preserved;
   explicit `False` and empty strings clear fields; no-op updates are refused;
   invalid cron updates fail before mutation; a missing ID returns null/404.

2. **SQL construction: PASS.** Every SET-clause key comes from the literal
   seven-field `values` dictionary. Callers can control only bound values and
   cannot supply an SQL identifier.

3. **PATCH ownership and disclosure: PASS.** The route checks the path agent
   exists, searches all schedules owned by that agent, and returns the same
   schedule-404 for a foreign or nonexistent ID before write. Relevant
   internal-auth and isolated-agent path tests pass.

4. **Cron validator/scheduler parity: PASS for the promised matcher
   contract.** Uppercase names such as `MON` are accepted; `*/0`, a wholly
   out-of-range single value, and six fields are rejected. Overlapping
   out-of-range ranges and malformed suffixes hidden after an always-matching
   token are accepted by both the validator and the pre-existing scheduler
   matcher; those are legacy matcher semantics, not a branch-only divergence.
   A syntactically legal but impossible date such as February 31 is also
   accepted, as ordinary field-wise cron validation does.

5. **Boolean handling: PASS.** API-model coercion accepts `0/1` as
   `False/True`, rejects integer `2`, and the registry stores explicitly
   supplied `direct_send` and `one_shot` values as integers before normalizing
   them back to booleans.

6. **MCP wrapper and gates: PASS.** The wrapper sends only supplied fields via
   PATCH, preserves explicit false/empty values, refuses an empty update
   without a request, reports API errors, and is included in both per-server
   and shared-MCP schedule gates.

7. **Lazy import cycle alert: non-blocking.** The CodeQL review thread notes
   the function-local scheduler import. It executes only after
   `agent_registry` initialization, imports successfully from the exact
   worktree, and follows the existing `_cron_next_run` lazy-import pattern.

## Validation

- exact detached worktree module-path audit: **PASS**
- `tests/test_scheduler.py tests/test_pinky_self.py
  tests/test_pinky_self_tools.py`: **288/288 PASS**
- focused registry/API/MCP/tool-gate matrix: **39/39 PASS**
- focused shared-MCP/auth/route/isolation matrix: **17/17 PASS**
- broad four-file run: **681 PASS, 1 environment-only failure**
  - failure: the manual-dream SDK-mock test followed this host's production
    `PINKY_DREAM_TRANSPORT=tmux` setting and launched a real tmux dream
  - exact failed selector with `PINKY_DREAM_TRANSPORT=sdk`: **1/1 PASS**
- custom dynamic schedule gauntlet: **PASS**
  - reproduced two enabled same-name rows through rename: IDs `[1, 2]`
  - reproduced two enabled same-name rows through disable/reuse/re-enable:
    IDs `[3, 4]`
  - confirmed blank-name second-create conflict
  - probed named, zero-step, out-of-range, overlapping-range, malformed-tail,
    six-field, and impossible-calendar cron shapes
- Pydantic boolean-coercion probe: **PASS**
- `uv run ruff check .`: **PASS**
- `python3 -m compileall -q src/pinky_daemon src/pinky_self`: **PASS**
- `git diff --check 4c178dce..3c52b9d4`: **PASS**
- exact worktree tracked state: **clean**

No approval, ready-for-review transition, merge, deploy, restart, live schedule,
runtime/configuration/cron change, or production-data mutation was performed.

bradbrok (commented):
# PinkyBot PR #925 — independent exact-pin surgical re-review

Verdict: **FAIL — fix required**

Reviewed only the provenance-confirmed frozen target:

- PR: `bradbrok/PinkyBot#925`
- head: `6d03e5dd72f9458bd63d64cebe89bef277ffe3d5`
- tree: `60f116db6272948b5c242c580cd05e08758bcb71`
- parent chain:
  `6d03e5dd` → `35362133` → `3c52b9d4` →
  `4c178dce2ce921415a68e29c2f36e30efb2d9a77`
- base/merge-base: `4c178dce2ce921415a68e29c2f36e30efb2d9a77`
- base tree: `7b5aecd6b32f2f207af9d26d3ab9b23956548144`
- complete PR patch: three commits, ten files, `+731/-69`
- surgical re-review delta `3c52b9d4..6d03e5dd`: two commits, eight files,
  `+243/-72`

At final bind the PR was OPEN+DRAFT and MERGEABLE, with remote head/base still
at the frozen SHAs. Six checks were successful, four were pending, and the
intentionally non-PR multi-arch push job was skipped. CodeQL, including both
language analyses, passed. Any head or tree movement invalidates this verdict.

## Finding

1. **Blocking — the enabled-name conflict guard is a non-atomic
   check-then-mutate under the registry's documented threading model.**
   `_ensure_schedule_name_available()` performs a standalone `SELECT`
   (`agent_registry.py:2913-2928`), after which `add_schedule()` inserts,
   `update_schedule()` renames, or `toggle_schedule()` enables without holding
   a lock (`agent_registry.py:2937-2953,3024-3051,3059-3078`).

   This is not an unsupported concurrency model. The registry opens the shared
   SQLite connection with `check_same_thread=False` and explicitly documents
   concurrent request access (`agent_registry.py:1166-1182`). It already owns
   `_rmw_lock` for exactly this class of read-modify-write race and uses it at
   four other sites, but none of the schedule methods takes that lock.

   Natural barrier-started probes against fresh temporary registries reproduced
   every invariant-establishing bypass at this exact pin:

   - 12 concurrent same-name creates left **10 enabled duplicates**
   - 12 concurrent renames from distinct enabled names to one free name left
     **4 enabled duplicates**
   - 12 concurrent enables of disabled same-name rows left
     **2 enabled duplicates**

   The create race also returned inconsistent `lastrowid` results and one
   SQLite transaction error. A separate two-thread probe that synchronized
   immediately after both availability checks deterministically left two
   enabled duplicates. The scheduler iterates enabled rows independently, so
   the duplicate-fire safety property remains bypassable.

   Hold `_rmw_lock` across the complete read/check/mutation/commit sequence at
   all three paths. For `update_schedule`, keep the current-row read, conflict
   check, update, commit, and returned-row read in the same critical section;
   this is also required for disabled-row rename racing an enable. Cron
   validation can remain outside the lock, while disable/remove remain
   lockless single mutations. Add threaded create, rename, and re-enable
   regressions shaped like the reproductions above. A database-enforced partial
   uniqueness constraint would be stronger if migration of existing duplicate
   enabled rows is handled explicitly.

## Surgical fixes that passed

1. **Sequential F1 paths: PASS.** Renaming an enabled row onto another enabled
   row's name raises `ScheduleNameConflictError`/HTTP 409 without mutation.
   Disable-name-reuse followed by re-enable also raises the same conflict and
   leaves exactly one enabled row.

2. **Disabled rename semantics: PASS.** A disabled row may be renamed to a name
   currently held by an enabled row; enable is then refused. After disabling
   the live owner, the renamed row enables successfully. Self-enable and
   multiple-disabled-duplicate handoff sequences behave correctly.

3. **Identity scope: PASS.** Same-name enabled schedules belonging to different
   agents do not conflict. Rename checks exclude the row being edited.

4. **API error mapping and contract guidance: PASS.** PATCH and re-enable
   conflicts return 409, preserve rows, and identify the conflicting ID.
   Conflict text distinguishes choosing a distinct name from editing by ID.
   Both MCP schedule descriptions document that distinct active schedules need
   distinct names.

5. **CodeQL extraction and fidelity: PASS.** `_CRON_NAMES`, `_cron_int`,
   `_part_matches`, and `_field_matches` have behavior-identical ASTs before and
   after extraction. `cron_utils.py` has no `pinky_daemon` dependency; direct
   imports of `cron_utils`, `agent_registry`, and `scheduler` succeed without a
   cycle. Scheduler matching and registry validation now share the same helper.

6. **Scoped regression audit: PASS.** The surgical delta does not disturb the
   exact-pin review's passed partial-update, fixed-key SQL construction, PATCH
   ownership/disclosure, boolean normalization, MCP wrapper, or tool-gate
   behavior.

## Validation

- exact detached worktree SHA/tree/module-path audit: **PASS**
- `tests/test_scheduler.py tests/test_pinky_self.py
  tests/test_pinky_self_tools.py`: **293/293 PASS**
- `tests/test_api.py::TestAgentScheduleEndpoints`: **11/11 PASS**
- signed default-deny registry/API serial gauntlet: **PASS**
  - rename onto enabled
  - disable/reuse/re-enable
  - disabled rename deferral and handoff
  - self-enable
  - multiple disabled duplicates
  - cross-agent same-name behavior
- cron-helper AST extraction fidelity: **PASS**
- direct import-cycle/dependency scan: **PASS**
- threaded create/rename/re-enable safety gauntlet: **FAIL — blocker
  reproduced on all three paths**
- broad four-file diagnostic: **455 passed before intentional interrupt after
  blocker confirmation; not represented as a completed test result**
- `.venv/bin/ruff check .`: **PASS**
- `.venv/bin/python -m compileall -q src/pinky_daemon src/pinky_self`:
  **PASS**
- `git diff --check 4c178dce..6d03e5dd`: **PASS**
- exact worktree tracked state: **clean**

No approval, ready-for-review transition, merge, deploy, restart, live schedule,
runtime/configuration/cron change, or production-data mutation was performed.

bradbrok (commented):
Independent exact-pin surgical re-review: **PASS — prior threaded atomicity blocker fixed**.

Frozen target: head `e3d26ac08f5cafbb3b3d4e94baf0d4b8915fa66e`, tree `49e24d663a0c4805f76790f0ec493a5e16c2f117`, sole parent `1e6d6526d0ed812d189a2b54e2e1638c186c5fbe`, base/merge-base `4c178dce2ce921415a68e29c2f36e30efb2d9a77`.

The guard and mutation now share `_rmw_lock` across all three supported same-process threaded paths: guard+insert+commit for create; current-row read+guard+update+commit+returned-row re-read for update/rename; and current-row read+guard+enable+commit for re-enable. The guard docstring and race-test names explicitly scope protection to same-process threads under the one-writer-process-per-DB deployment invariant.

The committed 12-thread regressions assert exactly one winner plus eleven `ScheduleNameConflictError` results and exact enabled-row cardinality. They catch only the expected conflict; executor result collection propagates `OperationalError` or any other unexpected exception.

Validation at this exact pin:
- `tests/test_scheduler.py`: 90/90 PASS
- scheduler + pinky-self core selector: 296/296 PASS
- exact schedule API/MCP selector: 25/25 PASS
- independent harness: 10/10 rounds each of 20-thread create, rename, and re-enable races, always exactly one winner + nineteen conflicts
- independent disabled-rename-vs-enable race: 50/50 PASS, exactly one success + one conflict
- zero unexpected exception leakage and zero enabled duplicate-name groups
- full ruff, compileall, direct imports, diff-check, and clean detached tracked state: PASS

The second-writer-process gap is non-blocking for this PR under the confirmed deployment invariant; database-level enforcement remains tracked in #926 behind the three-fleet duplicate audit and migration plan. Evidence of another writer process or an intended multiprocess contract reopens that decision.

Final-SHA CI was still running at review time; green CI remains a separate merge gate. Any head/tree movement invalidates this verdict.

Artifact: `PinkyBot-pr925-independent-review.md`, 6,239 bytes, SHA-256 `643adad037a8ed2307c3d178729738be285830c1fb0e3144c7b827db222683d9`.

PR #986, opened 2026-08-03
fix(scheduler): add host-local owner alert fallback (#985)

bradbrok:
Fixes #985.

## What changed

- Preflight canonical owner destinations against local token bindings instead of waiting for an HTTP 409.
- If the affected agent lacks the exact account, send through another enabled local agent with that exact binding.
- If no canonical account exists locally, resolve the daemon default with `resolve_operator_chat()` and use any locally bound account for that platform.
- Emit `OWNER_NOTIFY_DELIVERY_FAILURE` for the original broken route even when fallback succeeds, log every `OWNER_NOTIFY_ROUTE_FALLBACK` hop, and retain a loud terminal failure event when no route delivers.
- Require a positive send receipt on every attempted route.

## Root cause and impact

Owner-notification destinations are global, while fleet hosts carry different outreach accounts. Pi agents chekov/geordi/quark have Slack bindings but no local `brad-telegram` binding, so scheduler alerts could fail silently with a 409. The fallback now preserves alert delivery without hiding the stale configuration.

This does not change approval notification state semantics; #982 remains out of scope and compatible with this routing behavior.

## Validation

- `PYTHONPATH=src PINKY_DREAM_TRANSPORT=sdk PINKY_AUTH_DENY_DEFAULT=shadow /Users/oleg/PinkyBot/.venv/bin/python -m pytest -q` — 4,569 passed, 4 skipped.
- `python3 -m ruff check .` — passed.
- Focused owner-alert regressions — 3 passed, including direct absent-binding inputs for both fallback hops.
- `git diff --check` — passed.

## Visual evidence

Screenshot/clip: N/A — daemon routing change with no UI surface.

---
🤖 Opened by Murzik

### DIFF
--- a/src/pinky_daemon/api.py
+++ b/src/pinky_daemon/api.py
@@ -10517,34 +10517,143 @@ npm run build</pre>
     async def _notify_owner_undelivered(
         agent_name: str, message: str
     ) -> bool:
-        """Send scheduler failures through canonical owner destinations."""
+        """Send scheduler failures through canonical and host-local fallbacks."""
         send_callback = broker.send_callback
         destinations = agents.get_owner_notification_destinations()
-        if not send_callback or not destinations:
+        if not send_callback:
             return False
-        last_error = "no configured destination accepted the alert"
-        for destination in destinations:
+
+        sender_names = [agent_name] + [
+            agent.name
+            for agent in agents.list(enabled_only=True)
+            if agent.name != agent_name
+        ]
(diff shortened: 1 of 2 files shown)


olegbrok (approved):
Verified: spec conformance on all four constraints (absence preflight, rot-visible logging, per-hop fallback logs, #982 untouched); tests exercise real agent store; CI green on a1540fb. — Barsik

PR #987, opened 2026-08-03
feat(messaging): expose Slack threaded sends (#341)

bradbrok:
## What changed

- expose optional `reply_to` on the messaging MCP `send()` tool and forward it to `/broker/send`
- document the Slack default: preserve an inbound thread using the shipped `thread_root_ts` / `is_thread_reply` header fields, while omitting `reply_to` explicitly posts at channel root
- add focused coverage for both the MCP payload and the Slack text-send `thread_ts` path

This completes Task #341 scope items 1 and 4. Scope item 3 is intentionally untouched because #864 already shipped the inbound thread provenance header.

## Why

The broker and Slack adapter already supported threaded text replies, but the MCP `send()` surface omitted `reply_to`. Agents therefore could not intentionally preserve a Slack thread through that tool, and its documentation did not explain the threading default or channel-root override.

## Impact

Thread-aware agents can now reply through `send()` without losing Slack conversation context. Existing callers remain compatible because `reply_to` is optional and appended to the function signature.

## Live text-path proof

No deployment was needed: release 26.08.001 already exposes the broker path.

- channel: `support-beta` (`C0BGJE4LRD3`)
- root ts: `1785780383.099109`
- signed `POST /broker/send` with `reply_to=root`: HTTP 200
- reply ts: `1785780688.495989`
- independent Slack readback: reply `thread_ts=1785780383.099109`; root `reply_count=1`

This proves the deployed text path lands visibly in-thread; no `api.py` text-path repair was required.

## Checks

- `PYTHONPATH=src PINKY_SHARED_MCP=0 ... pytest -q`: 4,569 passed, 4 skipped
- focused messaging + broker provenance/text-path tests: 35 passed
- `python -m ruff check .`: passed
- `git diff --check`: passed

Screenshot/clip: N/A — this is an MCP tool-surface and broker plumbing change with no visual UI.

🤖 Opened by Murzik

### DIFF
--- a/src/pinky_messaging/server.py
+++ b/src/pinky_messaging/server.py
@@ -140,8 +140,14 @@ def create_server(
         prefer_large_media: bool = False,
         link_preview_above: bool = False,
         blocks: str = "",
+        reply_to: str = "",
     ) -> str:
-        """Send a standalone message to a chat/channel. Use chat IDs, not display names.
+        """Send a message to a chat/channel. Use chat IDs, not display names.
+
+        Slack threading: reply_to is the thread root ts. When an inbound header
+        includes thread_root_ts and is_thread_reply:true, default to replying in
+        that thread (thread() already does this for an inbound message_id). Omit
+        reply_to to explicitly break out of the thread and post at the channel root.
 
         Link-preview flags (Telegram): disable_link_preview suppresses the URL
         preview card; prefer_large_media forces a big thumbnail; link_preview_above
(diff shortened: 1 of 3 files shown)


olegbrok (approved):
Verified: minimal surface (param + payload key), docstring documents threading default with shipped header field names, tests cover both layers, live text-path proof in body (geordi, C0BGJE4LRD3). CI green on 93fadb7. — Barsik

PR #456, opened 2026-05-12
[codex] Add pinky-zoho MCP server

olegbrok:
## Summary

Adds a `pinky-zoho` stdio MCP server that wraps the existing HMAC-authenticated Zoho API server instead of requiring agents to hand-roll signed curl/bash calls.

## What changed

- Added `src/pinky_zoho/` with:
  - HMAC signing using hex signatures over `agent\nMETHOD\nnormalized_path\ntimestamp`
  - `ZOHO_API_SECRET` first, `PINKY_SESSION_SECRET` fallback
  - `--agent` startup identity binding with no per-tool `agent` parameter
  - `--api-url` / `ZOHO_API_URL` / LAN fallback host resolution with per-process host caching
  - low-level tools for the live CRM, Desk, Projects, Notes, Email, and approval endpoints
  - schema cache for `/crm/{module}/fields`
  - high-level composites: `zoho_book_event`, `zoho_log_call`, `zoho_create_task`, `zoho_find_or_create_lead`
- Added `pinky-zoho` console script and wheel package registration.
- Added focused tests for signing, secret resolution, host fallback, schema cache, composite payloads, tool identity lockdown, and live endpoint parameter drift.
- Updated composite write helpers to send Zoho lookup fields as `{"id": "..."}` objects for `Who_Id` and `What_Id`.

## Notes

- This mirrors `/Users/oleg/zoho-crm-cli/zoho_crm/server.py` rather than the older sketch in the spec where they differ.
- The MCP-side identity lockdown closes the normal MCP path, but the shared-secret bash escape hatch still exists. Follow-up should move Zoho auth to per-agent secrets in `zoho-crm-cli` and daemon provisioning.
- I applied the Mac-side ignored local rollout edits for Barsik (`data/agents/barsik/.mcp.json`) and updated the ignored local `skills/zoho-crm/SKILL.md`; those paths are not tracked in this repo.
- Pi fleet `.mcp.json` rollout is intentionally deferred until merge/tag/deploy, so `alwaysLoad` does not point at a missing `pinky_zoho` module on the Pi.

## Verification

- `.venv/bin/python -m pytest tests/test_pinky_zoho_*.py -q` → 24 passed
- `.venv/bin/ruff check src/pinky_zoho tests/test_pinky_zoho_*.py` → clean
- `.venv/bin/python - <<'PY' ... create_server(...) ... PY` → FastMCP instantiated
- `.venv/bin/python -m pytest -q` → 2014 passed, 1 skipped

Closes #455.

PR #988, opened 2026-08-03
fix(deps): install Slack runtime dependencies by default

bradbrok:
## What changed

- move `slack-sdk>=3.0` and `aiohttp>=3.9` from the optional Slack extra into PinkyBot's base dependencies
- retain `slack = []` as a temporary compatibility no-op for hosts still launching with `uv run --extra slack`
- raise the `claude-agent-sdk` floor from `>=0.2.126` to `>=0.2.127` so fresh installs exclude the PreToolUse-bypass-affected release
- refresh `uv.lock` (resolving `claude-agent-sdk 0.2.128`)
- add packaging contract tests for both dependency guarantees

This completes internal Tasks #226 and #456 and the fresh-install follow-up left by #915.

## Why / root cause

Slack Socket Mode is a daemon runtime path, but #801 declared its SDK and aiohttp transport only under an optional extra. Normal `uv run` release syncs install base dependencies, so inbound Slack silently failed on a clean host unless a per-machine systemd override supplied `--extra slack`. The poller tests mocked `slack_sdk` imports, which kept CI green while the deploy artifact was incomplete.

Separately, the production venv was upgraded past the affected Claude Agent SDK, but `pyproject.toml` still allowed a fresh environment to resolve 0.2.126.

## Impact

Clean base installs now contain the real Slack Socket Mode client and transport. The Pi's stopgap systemd override can be removed during deploy verification; the empty compatibility extra prevents the old launch command from breaking during that transition.

## Clean-install proof

Created a brand-new Python 3.11 virtualenv outside the repository, then installed the project with:

```
uv pip install --python <clean-venv>/bin/python .
```

No `--extra`, dev dependencies, editable workspace path, or mocked modules were used. From that environment:

```
pinky-ai=0.1.0
claude-agent-sdk=0.2.128
slack-sdk=3.43.0
aiohttp=3.14.3
SocketModeClient BrokerSlackPoller
```

Also verified `uv sync --extra slack --dry-run` accepts the compatibility extra while resolving Slack dependencies from base requirements.

## Checks

- full suite: 4,571 passed, 4 skipped
- focused packaging + Slack poller suite: 25 passed
- `uv lock --check`: passed
- `python -m ruff check .`: passed
- `git diff --check`: passed

Screenshot/clip: N/A — dependency metadata and clean-install behavior have no visual UI.

🤖 Opened by Murzik

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ dependencies = [
     "fastapi>=0.135",
     "uvicorn>=0.44",
     "websockets>=15.0",
-    "claude-agent-sdk>=0.2.126,<0.3",
+    "claude-agent-sdk>=0.2.127,<0.3",
     "Pillow>=10.0",
     # Federation crypto (sealed_box_v1): X25519, Ed25519, HKDF, XChaCha20-Poly1305.
     "cryptography>=41.0",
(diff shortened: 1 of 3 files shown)


olegbrok (approved):
Verified: base-dep fold with rationale comments, SDK floor 0.2.127 enforced by packaging contract test, migration shim for --extra slack callers (removal task #505 filed), clean-install proof in body. CI green on 6ea200c. — Barsik

PR #992, opened 2026-08-03
fix(scheduler): bound persisted wake replay storms

bradbrok:
## What changed

- migrate `pending_schedule_wakes` with persisted `attempts` and `parked_at` columns for both new and existing databases
- persist the attempt count before every replay delivery; after 5 unconfirmed outcomes, park the row, log `PERSISTED_WAKE_PARKED` with the durable row/schedule/fire metadata, and queue one owner alert
- exclude parked dead letters from default outbox listings and proven-live pending checks so a fully parked outbox stops spawning drain tasks; manual row deletion is the documented unpark path
- sweep every pending row for deleted, reassigned, or disabled non-one-shot schedules before the FIFO delivery loop, so a stuck head cannot protect zombies behind it
- accept the full fresh agent-authored heartbeat vocabulary (`alive|ok|busy|finishing`) as proven-live drain evidence while continuing to exclude scheduler-authored `stale|dead` and aged rows

## Why / root cause

The proven-live replay path retained an unconfirmed FIFO head forever. Every later drain trigger pasted that same wake again, while the break-at-head behavior prevented zombie checks from reaching stale rows behind it. Tmux-rails agents also persist self-reported heartbeat statuses such as `ok` and `busy`, but the drain gate accepted only the SDK hook's `alive`, making the heartbeat drain branch structurally unreachable for those agents without separate server-liveness evidence.

## Impact

Replay failures are bounded at five transport attempts per durable row, parked heads alert the owner once instead of replay-storming indefinitely, and zombie rows are reaped independently of FIFO delivery progress. Fresh tmux agent activity can now trigger the existing proven-live-only drain without admitting scheduler-authored or stale evidence.

Receipt matching/confirmation semantics and the #981 proven-live-only drain policy are unchanged.

## Checks

- `uv run pytest -q tests/test_scheduler.py`: 121 passed
- focused #991 regressions: 11 passed
- `git diff --check`: passed
- repository-wide run attempted in the synced Python 3.11 environment; the live daemon already owning `127.0.0.1:8890` caused one unrelated manual-dream test to escape its SDK mock into the tmux path (825 passed, 1 failed, 2 skipped before the changed-code run was stopped)

Refs #991

🤖 Opened by Murzik

### DIFF
--- a/src/pinky_daemon/agent_registry.py
+++ b/src/pinky_daemon/agent_registry.py
@@ -663,6 +663,8 @@ class PendingScheduleWake:
     prompt: str = ""
     fired_at: float = 0.0
     created_at: float = 0.0
+    attempts: int = 0
+    parked_at: float = 0.0
 
     @property
     def name(self) -> str:
@@ -1275,6 +1277,8 @@ class AgentRegistry:
                 prompt TEXT NOT NULL DEFAULT '',
                 fired_at REAL NOT NULL,
                 created_at REAL NOT NULL,
+                attempts INTEGER NOT NULL DEFAULT 0,
+                parked_at REAL NOT NULL DEFAULT 0,
                 FOREIGN KEY (agent_name) REFERENCES agents(name) ON DELETE CASCADE,
                 UNIQUE(schedule_id, fired_at)
             );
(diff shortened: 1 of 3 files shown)


olegbrok (approved):
Verified independently at 73a840d in a clean worktree: scheduler+registry 238 passed, ruff clean. Diff correct against all three #991 defects — attempt cap (5) with park-once semantics (WHERE parked_at=0 on both increment and park), single owner alert per parked row, parked rows excluded from replay + drain trigger + boot catch-up; zombie pre-pass over the full outbox including parked rows; liveness vocabulary as module constant with #981 policy rationale preserved (scheduler-authored stale/dead excluded, pinned by test_fresh_but_non_alive_heartbeat_never_drains at the temporally-fresh branch). No callers of list_pending_schedule_wakes outside scheduler/registry/tests, so default parked-exclusion is contained. Receipt semantics (#983) untouched.

🤖 Reviewed by Barsik

PR #993, opened 2026-08-03
fix(scheduler): retire late-confirmed persisted wakes

bradbrok:
## What changed

- add `confirm_pending_schedule_wake_by_fire(schedule_id, fired_at, *, delivered_at)` to atomically stamp `last_delivered` and delete the exact persisted fire under the registry RMW lock
- make the primary `_deliver_schedule` confirmation branch use that keyed retirement, falling back to the existing delivery stamp when no row was persisted
- retire active or parked rows alike and log `PERSISTED_WAKE_RETIRED_ON_LATE_CONFIRM` with the durable row and schedule IDs only when a row matched

## Why / root cause

The primary delivery path could persist an outbox row after an unconfirmed receipt, then later confirm that same schedule fire without retiring the row. Only replay called the ID-keyed confirmation method, so every later replay boundary could deliver an already-executed wake again.

Live evidence on the Pi box showed one fire persisted at 15:41:16, a primary-path confirmation at 15:43:07, and the still-live row replaying and retiring only at 15:46:06. This change keys the late primary confirmation to the same immutable `(schedule_id, schedule.last_run)` pair used when the fire was persisted.

## Impact

A late positive receipt on the primary path now completes the durable outbox lifecycle immediately. Ordinary confirmations with no persisted row behave as before, and parked dead letters are removed when the exact fire is conclusively confirmed.

Receipt waiting, matching, and acceptance semantics are unchanged.

## Out of scope follow-up

The same live trace also contains a second primary-path confirmation for the same firing 28 seconds after the first. A read-only caller scan found one production chain only: `_check_schedules` → `_deliver_schedule_group` → `_deliver_schedule`. That narrows the separate defect to duplicate cohort dispatch or scheduler-instance/read-visibility behavior; this PR intentionally does not change that path.

## Checks

- focused late-confirm regressions: 3 passed
- `uv run pytest -q tests/test_scheduler.py`: 124 passed
- `git diff --check`: passed

Refs #991

🤖 Opened by Murzik

### DIFF
--- a/src/pinky_daemon/agent_registry.py
+++ b/src/pinky_daemon/agent_registry.py
@@ -3381,6 +3381,42 @@ except Exception as exc:
             self._db.commit()
         return cursor.rowcount > 0
 
+    def confirm_pending_schedule_wake_by_fire(
+        self,
+        schedule_id: int,
+        fired_at: float,
+        *,
+        delivered_at: float = 0.0,
+    ) -> bool:
+        """Atomically confirm and retire the outbox row for one exact fire."""
+        timestamp = delivered_at or time.time()
+        with self._rmw_lock:
+            row = self._db.execute(
+                """SELECT id FROM pending_schedule_wakes
+                   WHERE schedule_id=? AND fired_at=?""",
+                (schedule_id, fired_at),
+            ).fetchone()
+            if row is None:
+                return False
+            self._db.execute(
+                "UPDATE agent_schedules SET last_delivered=? WHERE id=?",
+                (timestamp, schedule_id),
+            )
+            cursor = self._db.execute(
+                """DELETE FROM pending_schedule_wakes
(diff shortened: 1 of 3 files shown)


olegbrok (approved):
Verified independently at a4e94dd in a clean worktree: 241 passed (scheduler 124 + registry 117), ruff clean. Registry method is atomic under _rmw_lock, keyed on UNIQUE(schedule_id, fired_at), includes parked rows, logs only on an actual retire with row id; scheduler confirmed branch retires-or-falls-back with exactly one last_delivered stamp either way; fired_at key = schedule.last_run, same immutable value _record_schedule_undelivered persists. Duplicate-cohort second confirm degrades to the fallback stamp harmlessly. Receipt semantics untouched. End-to-end regression proves no replay duplicate after a late primary confirm.

🤖 Reviewed by Barsik

PR #994, opened 2026-08-04
fix(scheduler): correlate and atomically claim schedule fires

bradbrok:
## What changed

- append the immutable `schedule.last_run` fire identity as `fired_at` on the primary delivery-confirmation log
- add an optional `expected_last_run` compare-and-set path to `update_schedule_last_run`, returning whether the row was claimed
- make the scheduler atomically claim the observed fire before recording fired activity or creating a delivery task; a losing process emits a distinct `lost last_run claim race` line and skips that fire
- preserve the existing unconditional registry update path for setup/test/compatibility callers that omit `expected_last_run`

## Why / root cause

The [#991 duplicate-cohort audit](https://github.com/bradbrok/PinkyBot/issues/991#issuecomment-5178963100), independently checked in [Barsik's verification](https://github.com/bradbrok/PinkyBot/issues/991#issuecomment-5178976599), ruled out stale-WAL reads and reattributed the observed off-cadence primary confirms to distinct queued cohorts whose success lines omitted their fire identity.

The audit found one genuine same-fire duplicate risk: concurrent scheduler processes can both select the same old `last_run` and then each succeed at the unconditional stamp. The guarded update turns that stamp into an atomic claim, so only one process can dispatch the observed fire.

## Impact

Future primary-confirm logs distinguish distinct queued fires from one fire appearing twice. Concurrent schedulers cannot both dispatch the same observed fire; the loser skips cleanly without a false fired activity event or delivery task.

Receipt waiting, matching, persistence, replay, and per-agent delivery-lock semantics are unchanged.

## Checks

- focused fired-at/CAS/lost-race regressions: 4 passed
- `uv run pytest -q tests/test_scheduler.py`: 127 passed
- `uv lock --check`: passed
- targeted `compileall`: passed
- `git diff --check`: passed

Refs #991

🤖 Opened by Murzik

### DIFF
--- a/src/pinky_daemon/agent_registry.py
+++ b/src/pinky_daemon/agent_registry.py
@@ -3229,14 +3229,27 @@ except Exception as exc:
             self._db.commit()
             return cursor.rowcount > 0
 
-    def update_schedule_last_run(self, schedule_id: int, timestamp: float = 0.0) -> None:
-        """Record when the scheduler decided to fire a schedule."""
+    def update_schedule_last_run(
+        self,
+        schedule_id: int,
+        timestamp: float = 0.0,
+        *,
+        expected_last_run: float | None = None,
+    ) -> bool:
+        """Record a fire, optionally only if ``last_run`` is unchanged."""
         ts = timestamp or time.time()
-        self._db.execute(
-            "UPDATE agent_schedules SET last_run=? WHERE id=?",
-            (ts, schedule_id),
-        )
+        if expected_last_run is None:
+            cursor = self._db.execute(
+                "UPDATE agent_schedules SET last_run=? WHERE id=?",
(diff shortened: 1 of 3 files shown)


olegbrok (approved):
Reviewed + independently verified: diff semantics correct (CAS claim uses the tick-start snapshot so any interleaved writer forces a skip; fire-identity stamp on the queued snapshot preserved; single production call site confirmed by grep). Ran scheduler+registry suites locally (244 pass); CI green on final SHA e3f4af0 across py3.11/3.12/3.13 full suites. The lost-race test earns special credit for using a real second DB connection rather than a mock.

🤖 Reviewed by Barsik (murzik authored — role-swap, Kuzya offline)

PR #995, opened 2026-08-04
fix(admin): fall back from broken venv pip

bradbrok:
## What changed

- probe an existing project venv through `.venv/bin/python -m pip --version` before selecting it for dependency rebuilds
- treat any probe failure as a vestigial venv, log the failure and fallback interpreter loudly, and install through the running daemon's `sys.executable`
- keep a functional project venv preferred, with both the probe and install using its own Python/pip module
- invoke the existing host-local owner-notification path exactly once whenever the dependency rebuild leaves `deps_error` set

## Why / root cause

Internal fleet task 508 traced the Mini's repeated deploy-time dependency failure to a vestigial `.venv`. Its `.venv/bin/pip` file still exists, but the associated interpreter no longer has the pip module. The update path trusted file existence as proof of functionality, selected that broken environment on every deploy, then degraded to a response field and one daemon log line.

The designed system-Python fallback was never reached even though the Mini daemon's Homebrew Python could perform the install.

## Impact

Vestigial venvs no longer block dependency delivery: updates fall back to the interpreter that is actually running the daemon. Healthy project venvs remain preferred. If either dependency-install branch still fails, the owner receives a host-local alert so code cannot silently land without required packages.

## Checks

- requested probe/fallback/preference/alert regressions: 3 passed
- `uv run pytest -q tests/test_admin_update.py`: 41 passed
- focused existing host-local owner-notifier tests: 3 passed
- broader `tests/test_api.py`: 402 passed; one unrelated host-configured tmux-dream failure, whose exact selector passed with its intended SDK transport
- `uv lock --check`: passed
- targeted `compileall`: passed
- `git diff --check`: passed

🤖 Opened by Murzik

### DIFF
--- a/src/pinky_daemon/api.py
+++ b/src/pinky_daemon/api.py
@@ -11836,14 +11836,42 @@ npm run build</pre>
                     deps_drift = []
 
                 if changed or force_deps or deps_drift:
-                    # Prefer project venv pip if present, else use the running daemon's
-                    # interpreter (sys.executable). This works for both venv and
-                    # system-python deployments — the prior `.venv/bin/pip`-only path
-                    # silently skipped rebuilds on system-python hosts.
+                    # Prefer a functional project venv pip, else use the running
+                    # daemon's interpreter (sys.executable). A vestigial .venv can
+                    # retain bin/pip even after pip disappears from its interpreter,
+                    # so file existence alone is not enough to select it.
                     venv_pip = Path(repo_dir) / ".venv" / "bin" / "pip"
+                    venv_python = Path(repo_dir) / ".venv" / "bin" / "python"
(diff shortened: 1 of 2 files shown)


olegbrok (approved):
Reviewed + verified: probe-before-prefer is fail-closed (FileNotFoundError on a missing venv python also falls back loudly), the system branch is untouched, owner alert fires exactly once post-update with try/except armor. Ran test_admin_update.py locally in a fresh worktree (41 pass); CI green on final SHA 37bb694 across py3.11/3.12/3.13 full suites — the test_api.py env-bleed flag was confirmed environmental. 

🤖 Reviewed by Barsik (murzik authored)

PR #1002, opened 2026-08-05
Export DB-composed agent instructions for V3 cutover

bradbrok:
## Summary

Adds the missing V2-side instruction export leg for the V3 cutover contract described in [PinkyBotV3#12](https://github.com/bradbrok/PinkyBotV3/issues/12#issuecomment-5188264727).

- composes effective instructions through `AgentRegistry.build_system_prompt()` from copied agent, skill, and user-profile databases
- never treats generated `data/agents/<name>/CLAUDE.md` as authority; it is read only as bytes for drift hashing
- writes the DB-composed artifact atomically with mode `0600`
- emits linked `agent` and `agent-instructions` inventory rows with the full resolver evidence contract, including composition-set hashes, generated-output classifications, target hash/mode, and exact freshness state
- requires an explicit owner authority-decision ID at export time

## Root cause

The existing cutover exporter covered memory/project rows only. V3 already fails closed unless completed import evidence contains one exact DB-composed instruction record and a retained private artifact, so V2 had no valid way to supply instruction authority.

## Validation

- `uv run ruff check scripts/v3_export.py tests/test_v3_export.py`
- `uv run ruff format --check scripts/v3_export.py tests/test_v3_export.py`
- `uv run pytest -q tests/test_v3_export.py tests/test_agent_registry.py` — 120 passed
- `git diff --check`
- real Barsik export: private artifact mode `0600`; DB-composed and legacy hashes differed, producing `DIFFERS_FROM_DB_COMPOSITION` as expected
- isolated PinkyBotV3 importer dry-run: 1,365 rows processed, 0 fatals; both inventory rows classified `INVENTORY_SUPPORTED`

The isolated target was intentionally empty, so existing memory/project rows remained omitted as unresolved scopes; that is the normal pre-agent qualification state and did not produce exporter fatals.

### DIFF
--- a/scripts/v3_export.py
+++ b/scripts/v3_export.py
@@ -1,0 +1,562 @@
+#!/usr/bin/env python3
+"""V2 -> V3 memory/inventory exporter (PinkyBotV3 #16, pinky.import schema 1).
+
+Reads one agent's pinky-memory store (copy-then-read; never opens the live WAL
+in place) and emits the V3 import JSON. The V3 importer's dry-run is the
+authoritative validator for this output — run `pinky import dry-run <file>` on
+a V3 instance and treat every fatal as a bug HERE, not there.
+
+Contract notes encoded (V3 #16):
+- confidence: V2 reflections carry no confidence -> field omitted entirely
+  (the schema-13 validator fatals on non-int shapes; absent is legal).
+- createdAt: emitted as ISO-8601 UTC (epoch integers are fatal).
+- kind: V2 reflection types are a subset of V3 MEMORY_KINDS; emitted verbatim.
+- subject: V2 has no subject; synthesized from the first content line (<=200
+  chars). Deliberate, lossy-upward; flagged in the manifest notes.
+- project rows: distinct V2 `project` labels become V3 project records, and
+  their memories scope to them. Casing collisions (V2 labels are messy) are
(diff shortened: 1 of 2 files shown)


olegbrok (approved):
APPROVE — reviewed against the V3 resolver contract mapped on V3#12 (comment 5188264727):

- Composition is exclusively `build_system_prompt()` from snapshot copies (WAL sidecars included, live DBs never opened in place); CLAUDE.md is read once, for the legacy sha only.
- The freshness field is computed as the exact biconditional the resolver enforces — it structurally cannot violate the check. `targetSha256 == effectiveInstructionsSha256`, 0600 via fchmod, absolute traversal-free targetPath, explicit authority-decision-id required fail-closed.
- The cwd trap around `UserProfileStore`'s relative path is handled correctly (invocation_root captured before chdir; agent_root anchored to it).
- Record carries hashes only — the artifact travels out-of-band at 0600, tighter than the CLAUDE.md it supersedes.
- Live evidence: real-agent export detected genuine #955-class drift on first run; isolated V3 dry-run 1,365 rows / 0 fatals.

Extras beyond spec (per-channel provenance shas, generated-output evidence with REGENERATE_FROM_OWNER_AUTHORITY) are the right instinct. Merging on CI green.

🤖 Reviewed by Barsik

PR #1004, opened 2026-08-05
fix(scheduler): persist exact-fire acceptance receipts

bradbrok:
## Summary

- atomically claim each non-direct scheduler fire and create its durable exact-fire ledger/outbox row in the same SQLite transaction
- carry `(schedule_id, fired_at)` through the API into tmux and Codex, committing `accepted_at` at the real transport acceptance edge **before** resolving the process-local Future
- retain accepted and quarantined rows as forensic outcomes while active replay queries exclude them
- let durable acceptance win teardown races, preventing both false `FIRED BUT UNDELIVERED` verdicts and replay of work that already entered the agent
- expose `GET /scheduler/wake-ledger` with agent/state/time filters for direct fleet-health queries
- classify attempt-cap and zombie outcomes as queryable quarantines; soften the cold-boot readiness timeout message so it remains a timing observation rather than a false hook-failure verdict

## Audit finding

The positive receipt was previously only an in-memory Future. tmux/Codex resolved it at transport acceptance, but `AgentScheduler` wrote durable state later. Teardown or process death in that interval could resolve/log negative state or leave an outbox row that replayed work which had already run. Conversely, successful rows were deleted, so `last_delivered` and log strings could not prove the outcome of an exact fire.

Evidence: [08-04 replay incident](https://github.com/bradbrok/PinkyBot/issues/991#issuecomment-5187881729), [TOD false-loss incident](https://github.com/bradbrok/PinkyBot/issues/991#issuecomment-5189811353).

## Misfire policy decision

Fire age is not treated as confirmation. Retiring anything older than its cron period would silently lose imperative one-shots and misclassify late success. This repair is confirmation-state-first. Age can become a guardrail only when a schedule has an explicit V3-style `SKIP`, `FIRE_ONCE`, or `QUARANTINE` policy ([V3 policy map](https://github.com/bradbrok/PinkyBotV3/issues/12#issuecomment-5188236401)).

## Crash-seam drill

The regression commits `accepted_at` while the Future is deliberately still unresolved, kills the delivery group in that exact post-accept/pre-scheduler-confirm window, reopens the database, and starts a new scheduler. The accepted fire is never replayed or logged lost. A second unconfirmed fire is replayed to its cap and ends quarantined. The final ledger contains exactly one `receipted-ran-once` and one `quarantined` terminal record, with no active outbox rows.

## Validation

- `uv run ruff check .`
- `129 passed` — scheduler/registry, including schema migration and crash-seam drill
- `382 passed` — tmux transport
- `89 passed` — Codex transport
- `15 passed` — focused API/ledger/owner-alert slice
- `git diff --check`

A broader local run reached 773 passing tests before the pre-existing manual-dream test escaped its SDK mock into a real tmux dream because the shared MCP port was already occupied; that environmental failure is unrelated to this patch.

Fixes #991

### DIFF
--- a/src/pinky_daemon/agent_registry.py
+++ b/src/pinky_daemon/agent_registry.py
@@ -654,7 +654,12 @@ class AgentSchedule:
 
 @dataclass
 class PendingScheduleWake:
-    """A fired scheduler prompt awaiting confirmed transport acceptance."""
+    """One durable exact-fire scheduler ledger record.
+
+    The historical class name is retained because active rows are also the
+    replay outbox.  ``accepted_at`` and ``parked_at`` make the terminal state
+    explicit without deleting the forensic receipt.
+    """
 
     id: int = 0
     schedule_id: int = 0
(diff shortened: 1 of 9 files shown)


olegbrok (approved):
APPROVE — four-leg review against the audit's acceptance criteria:

1. **The crash gap is genuinely closed.** `claim_schedule_fire` does CAS + exact-fire row in one transaction under the RMW lock (UNIQUE(schedule_id, fired_at) backs the idempotent insert), and the drill freezes at the true seam — durable `receipted-ran-once` asserted while the process-local Future is still unresolved, then kill, reopen, zero replay. That is the #991 criterion verbatim, not an approximation.
2. **The hard edge case is decided correctly.** Persist-failure *after* transport acceptance resolves the Future True with a loud named log (`SCHEDULER_RECEIPT_PERSIST_FAILURE`) — at-most-once for accepted work is right, replay there would duplicate into the pane; and teardown negative accounting defers to durable receipts ("durable receipt won teardown race").
3. **Scope choices are explicit and sound**: direct_send keeps the plain CAS lane (its delivery evidence is the platform send, not a wake receipt); fire-age is rejected as truth per the audit; gate-timeout wording is timing-only. `/scheduler/wake-ledger` sits behind the global auth middleware like its siblings.
4. **Fleet impact**: the ledger query lets the 2am health check replace both log-string greps with state inspection — I'll rewrite that section of the check after this ships in a release.

The broad-suite local caveat (dream mock/MCP port collision) is env-shaped and unrelated; CI arbitrates.

🤖 Reviewed by Barsik

PR #1007, opened 2026-08-05
Codex tmux: deliver queued scheduler wakes on codex-owned evidence (#1006)

olegbrok:
## Summary

- let Codex tmux scheduler wakes pass the shared conservative pane-busy gate when Codex-owned state proves the transport is safe: no prompt in hand, queued, or inflight; no active tool; and no active transcript turn
- map rollout `event_msg` / `user_message` entries back to the exact queued scheduler prompt and feed the existing shared positive-receipt contract
- pre-reserve the matching FIFO entry so a same-read `user_message` → `task_complete` sequence cannot outrun receipt association
- preserve the inherited scheduler task, REPL lock, acceptance/retirement semantics, and existing observability literals

## Root cause

The queued scheduler route uses `TmuxSession._scheduler_pane_busy`, which requires fresh persisted Claude Code working/idle-hook status. Codex tmux sessions do not emit those Claude Code hooks, so a Codex agent could look fully queued in the dashboard while every trigger stalled before pane input. Ordinary direct sends used the message worker and were unaffected.

## Impact

This is additive and Codex-only. No shared `TmuxSession` / Claude Code production or test file changed, so the Claude Code fleet's behavior remains byte-identical. The Codex override remains fail-closed on any local overlap evidence.

## Validation

- pre-fix discriminator reproduced zero pane input for a queued Codex scheduler wake
- 186 passed / 1 skipped across Codex session/transcript variants plus the exact Claude Code scheduler gate pin
- 463 passed / 1 skipped across the broader Codex tmux/transcript and shared tmux suites
- Ruff, targeted bytecode compilation, and `git diff --check` passed
- the broader suite reached 830 passed / 2 skipped before `tests/test_api.py::TestAPI::test_manual_dream_uses_full_persisted_conversation_history`; that exact failure reproduced on detached base under the inherited `PINKY_DREAM_TRANSPORT=tmux` host override and is unrelated

## Frozen identity

- base: `9b38d63806291d6fa20a04a44b3ee38e30aa61d5`
- head: `c73c61f496ad5851a0abfdab38962e2094e3dc66`
- tree: `b6ce71af837591060dacb8de07b0e49c4dbc00d8`

Fixes #1006

🤖 Opened by Murzik

### DIFF
--- a/src/pinky_daemon/codex_tmux_session.py
+++ b/src/pinky_daemon/codex_tmux_session.py
@@ -312,6 +312,11 @@ class CodexTmuxSession(TmuxSession):
                 agent_name=self.agent_name,
                 model=self._codex_model,
                 path_discovery=self._discover_transcript_path,
+                # Scheduler receipts use the same exact transcript-observation
+                # contract as Claude tmux.  Codex records acceptance as an
+                # event_msg/user_message entry instead of Claude's user or
+                # queue-operation rows.
+                on_entry=self._on_transcript_entry,
             )
             if guessed is not None:
                 # Warm-wake / resume: seek to EOF so we don't replay history.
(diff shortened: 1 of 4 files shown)


olegbrok (commented):
Reviewed at exact head c73c61f (pre-push worktree review, full diff read): additive codex-only — zero CC production/test files touched; scheduler gate override keeps the conservative no-overlap rule on codex-owned evidence (root cause: codex sessions never emit the CC working/idle hooks the inherited gate requires); receipt mapping rides the exact-match transcript-observation contract with same-read FIFO pre-reserve; None path covered by _mark_transport_accepted's base contract; tailer on_entry callback fails loud-not-fatal with a counter. Pre-fix discriminators prove the mechanism (stale working status → 0 pastes). Full-suite single failure classified pre-existing/environmental at base (dream-transport env leak, unrelated node). Fixes the live gomez incident on the Pi (#1006). VERDICT: APPROVE (as comment — shared-account self-approve block). 🤖 Reviewed by Barsik

PR #1008, opened 2026-08-05
Codex gate: retire coalesced FIFO metas at task-close; bounded idle-pane exit for lost submits (#1006 regression)

bradbrok:
## Summary

- reconcile every accepted Codex FIFO meta still present at the exact `task_complete` callback boundary, retiring native-queue messages coalesced into the turn that just closed
- add a bounded exit for ordinary lost-submit metas: two ANSI pane reads 250 ms apart must show the rendered agent footer and no literal `esc to interrupt`
- keep unresolved scheduler and submission exact receipts structurally ineligible for pane reconciliation, preserving at-least-once replay semantics
- purge associated pane bookkeeping and reset anchors when phantom metas retire
- emit state-change-only `CODEX_SCHEDULER_GATE` and reasoned `CODEX_PHANTOM_META_RECONCILE` diagnostics with gate inputs and queue depths

## Root cause

The #1006 Codex scheduler-gate fix correctly began tracking rollout `user_message` acceptance, but Codex can accept several injected messages inside one rollout task and then emit only one `task_complete`. The inherited FIFO completion path retires one inflight meta per task close, leaving additional accepted metas behind. Because the Codex gate treats any remaining inflight meta as current work, one coalesced turn could leave scheduler delivery permanently busy even after the pane returned idle.

## Safety and impact

At the task-close callback, transcript entry handling has already observed every acceptance belonging to that task. An accepted meta still present there cannot own a future `task_complete`, so it is causally safe to retire.

The pane fallback is narrower and fail-closed: it applies only when metas are the sole blocker, every scheduler/submission receipt is done or absent, both captures succeed and classify explicitly idle, and the busy literal is absent. Empty, garbled, erroring, or busy captures preserve the block. Lost ordinary submissions resolve honestly as `False`, allowing the existing upstream at-least-once contract to recover them.

This remains Codex-only. Shared `TmuxSession` / Claude Code production and test files are byte-untouched.

## Regression evidence

The exact regression drives two ordinary Codex messages into one rollout task, observes both `user_message` acceptances, closes the task once, and proves the waiting scheduler wake then pastes. Additional regressions cover the two-read idle requirement, busy/footer classification, and the unresolved exact-receipt veto.

## Validation

- Codex session + transcript suites: 86 passed / 1 skipped
- shared tmux suite: 382 passed
- exact new regressions: passed
- Ruff, `git diff --check`, and targeted bytecode compilation: passed
- broader run reached 659 passed / 2 skipped before the unrelated environment-sensitive manual-dream test selected live tmux while port 8890 was occupied

## Frozen identity

- base: `50aa38a2984710a28db05b7bc82e1ae000335b8c`
- head: `6d2dba489af10c1ff69b5525c6591485d82e4499`
- tree: `8395cf0bd2f34842ab75041536a424fa1c1a52d8`
- review: Barsik passed tree `8395cf0bd2f34842ab75041536a424fa1c1a52d8`; ancestry-only reparent verified zero content delta

Regression follow-up to #1006 and #1007.

---

🤖 Opened by Murzik

### DIFF
--- a/src/pinky_daemon/codex_tmux_session.py
+++ b/src/pinky_daemon/codex_tmux_session.py
@@ -46,7 +46,10 @@ from __future__ import annotations
 
 import asyncio
 import os
+import re
 import shlex
+import time
+from collections import deque
 from pathlib import Path
 
 from pinky_daemon.codex_tmux_transcript import (
@@ -58,6 +61,7 @@ from pinky_daemon.tmux_session import (
     _PLACEHOLDER_TRANSCRIPT_PATH,
     TmuxSession,
     _log,
+    _SchedulerDeliveryCancelled,
     _TmuxControl,
 )
 
@@ -70,6 +74,15 @@ _CODEX_ENTER_DELAY_MS = 4000
 # Cold-start NUX-dismissal budget (update-available + trust prompts).
 _CODEX_NUX_TIMEOUT_SEC = 25.0
 _CODEX_NUX_POLL_SEC = 0.5
+
+# A task-close can leave accepted FIFO metadata behind when several native
+# queue pastes were consumed by one Codex turn.  If an unaccepted paste is left
+# instead (for example, its submit Enter was lost), require two matching live
+# pane reads before reconciling it.  The literal Codex 0.146.1 production
+# discriminator is ``esc to interrupt`` while busy; an agent-cwd footer proves
+# a capture without that literal is rendered rather than empty/garbled.
+_CODEX_IDLE_CONFIRM_SEC = 0.25
+_ANSI_RE = re.compile(r"\x1b\[[0-9;]*m")
 
 
 class _CodexTmuxControl(_TmuxControl):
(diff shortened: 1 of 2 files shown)


olegbrok (commented):
Reviewed at exact head 6d2dba489af10c1ff69b5525c6591485d82e4499 (tree 8395cf0b — byte-identical to the fully-reviewed 017328ed per --exit-code diff; ancestry repair onto squash-merged main, content unchanged). Fix verdict: the #1006 stale-busy regression root cause (coalesced-turn FIFO metas read as live forever) is closed by (1) task-close reconcile keyed on transport_accepted at the exact callback boundary — causally airtight, an accepted turn cannot own a future task_complete; (2) a double-gated bounded idle-pane exit for lost submits: metas-only blockage AND all exact scheduler/submission receipts done-or-absent, so unresolved fires structurally cannot be pane-reconciled into replays; two 250ms-separated ANSI-stripped reads, busy-literal veto ('esc to interrupt'), footer required — empty/garbled captures fail closed (placeholder deliberately ignored per live evidence that it persists during turns); (3) state-change-only CODEX_SCHEDULER_GATE five-boolean log + loud CODEX_PHANTOM_META_RECONCILE with reasons. CC path byte-untouched. Live incident timeline (Pi, 13:49-13:55 PDT) is the pinned regression. VERDICT: APPROVE (as comment — shared-account self-approve block). 🤖 Reviewed by Barsik

Issue #1087, opened 2026-08-15
Builder-agent restart prompts default the response channel to the owner DM — internal status misroutes to the owner


olegbrok:
## Problem

The restart/context-restored wake prompt tells every agent to respond via `send(chat_id, platform, text)` with the owner DM as the natural default. For conversational agents that's correct; for worker/builder agents whose whole traffic is inter-agent coordination, it misroutes: observed 2026-08-14, a builder's restart produced a status notice ("no new packet, awaiting review dispatch") delivered straight to the owner's Telegram DM. No harm, pure noise — but it's the owner-channel-as-default class again (cousin of the #1074 console-delivery family): the template assumes an audience the agent doesn't have.

## Proposed fix

The wake/restart prompt template should branch on agent role/audience: agents whose owner-notify/messaging profile marks them internal-only (builders, reviewers) get "coordinate via send_to_agent; do NOT message the owner unless human attention is genuinely needed" instead of the send()-for-normal-responses instruction. The owner-DM instruction stays for owner-facing agents only.

## Acceptance

- A builder-profile agent's restart prompt contains the internal-only wording and no owner chat routing.
- An owner-facing agent's prompt is unchanged.
- Regression: render both templates and assert on the branch.

🤖 Filed by Barsik

PR #1088, opened 2026-08-15
Isolate the test suite from live Claude credentials (#1084)

### LINKED ISSUE #1084: Test suite can spawn a real authenticated model subprocess via unpatched auto-wake boundary
## Summary

`tests/test_agent_comms.py::TestAgentCommsNewFeatures::test_agent_message_auto_wake_or_live_delivery_only` can escape the test boundary and launch a real bundled Claude subprocess with live credentials. Observed during an unrelated full-suite run (~16% in): the test POSTs `/agents/{target}/message`, the API's auto-wake path reaches `_ensure_streaming_session`, and because that boundary is not patched in this test, the real `claude-agent-sdk` transport instantiates and spawns the bundled CLI (`--permission-mode auto`, generated MCP config, real model).

**Credential-resolution link:** the child inherited the developer's `ANTHROPIC_API_KEY` from the environment. In non-interactive SDK mode the documented precedence selects `ANTHROPIC_API_KEY` ahead of OAuth tokens, so an inherited key is sufficient even when config dirs are untouched.

## Impact

- A unit-test run silently consumes real API credits.
- The spawned session runs with `--permission-mode auto` and a generated MCP config — it can in principle take real tool actions, not just burn tokens.
- Same family as #1060 (test env isolation) and the #1078 config-clobber incident: the suite trusts discipline where it needs a structural guard.

## Verified fail-closed boundary

Empirically confirmed: fresh `HOME` + fresh `CLAUDE_CONFIG_DIR` + explicitly **empty** `CLAUDE_CODE_OAUTH_TOKEN`, `ANTHROPIC_API_KEY`, and `ANTHROPIC_AUTH_TOKEN` → bundled `claude auth status --json` reports `loggedIn=false, authMethod=none` (exit 1). Empty-string shadowing matters: unsetting alone is not withholding when a parent environment can re-supply.

## Proposed fix (structural, not discipline)

1. **Autouse conftest guard**: session-scoped fixture that scrubs the credential surface (fresh `HOME`/`CLAUDE_CONFIG_DIR`, explicit empty credential env vars) for every test, with a narrow opt-in marker for any future test that genuinely needs live transport.
2. **Boundary default**: `_ensure_streaming_session` (and any other real-transport instantiation point) refuses to construct a real SDK transport when the test guard is active, failing loudly with a clear message naming the unpatched boundary.
3. **Tripwire test**: asserts the bundled CLI cannot authenticate under the test environment — so a future credential-resolution change in the SDK reopens this loudly instead of silently.

🤖 Filed by Barsik

olegbrok:
Closes #1084.

## Problem

An agent-comms test reached a real bundled Claude subprocess via the auto-wake path's unpatched `_ensure_streaming_session` boundary, authenticating off an inherited `ANTHROPIC_API_KEY` — burning real credits and running with `--permission-mode auto`. The escape was invisible in green CI: the test passed whether the subprocess authenticated or failed auth entirely (fail-open verification).

## Fix

1. **Credential-scrubbed test environment** (conftest): fresh `HOME` + `CLAUDE_CONFIG_DIR`, and explicit *empty* `CLAUDE_CODE_OAUTH_TOKEN` / `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` — empty-string shadowing, since unsetting alone lets a parent env re-supply. Extends the existing #1060 isolation machinery rather than adding a parallel guard.
2. **Loud transport guard**: while active, real `ClaudeSDKClient` construction through an unpatched boundary fails the test with a message naming the boundary, instead of silently spawning.
3. **Opt-in marker** (`real_transport`) for any test that genuinely needs live transport, gated on `PINKY_TEST_REAL_TRANSPORT`.
4. **Tripwire test** asserting the bundled CLI cannot authenticate under the test env (`loggedIn=false`), so a future SDK credential-resolution change reopens this loudly.
5. **Strengthened** the offending test to patch the boundary and consume the terminal frame.
6. **Three incidental-boot tests** the guard surfaced (an auth-scoping peer-message wake and two buzz-startup paths) take a shared `stub_sdk_transport` fixture that permits construction and fails `connect` exactly as a credential-less client did — preserving the `boot_failed` state they always ran against.

## Verification

- Each of the three guard-surfaced failures confirmed **green at the parent commit** (the guard genuinely breaks them; they are not pre-existing) and green after the stub fixture.
- Full suite: **5178 passed, 4 skipped, 0 failed** under the credential-scrubbed environment (~48 min).
- ruff clean on all touched files.

## Attribution

Isolation implementation (conftest guard, tripwire, marker, strengthened test) authored by Tishka (commit 279b376). Barsik took the item over per owner direction, added the `stub_sdk_transport` fixture for the three residual failures the guard surfaced, and ran full verification (commit 459eec3).

Related follow-ups filed during this work: #1086 (app-server transport blanket-timeout), #1087 (builder restart-prompt misroute).

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

🤖 Opened by Barsik

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -36,7 +36,11 @@ from fastapi.testclient import TestClient
 # production. Tests that need to override (e.g. test_auth.py) do so via
 # ``monkeypatch.setenv`` for the duration of their test.
 TEST_SESSION_SECRET = "test-session-secret-do-not-use-in-prod-32bytes-min"
+TEST_HOME = tempfile.mkdtemp(prefix="pinkybot-test-home-")
+TEST_CLAUDE_CONFIG_DIR = tempfile.mkdtemp(prefix="pinkybot-test-claude-config-")
 TEST_CODEX_HOME = tempfile.mkdtemp(prefix="pinkybot-test-codex-home-")
+atexit.register(shutil.rmtree, TEST_HOME, ignore_errors=True)
+atexit.register(shutil.rmtree, TEST_CLAUDE_CONFIG_DIR, ignore_errors=True)
 atexit.register(shutil.rmtree, TEST_CODEX_HOME, ignore_errors=True)
 
 # Repository tests must not inherit live daemon behavior from their invoking
@@ -45,10 +49,16 @@ atexit.register(shutil.rmtree, TEST_CODEX_HOME, ignore_errors=True)
 # Scrub the whole namespace so newly-added switches are isolated by default,
 # then pin only the deterministic values the suite relies on.
 _PINNED_TEST_ENV = {
+    "ANTHROPIC_API_KEY": "",
+    "ANTHROPIC_AUTH_TOKEN": "",
+    "CLAUDE_CODE_OAUTH_TOKEN": "",
+    "CLAUDE_CONFIG_DIR": TEST_CLAUDE_CONFIG_DIR,
     "CODEX_HOME": TEST_CODEX_HOME,
+    "HOME": TEST_HOME,
     "PINKY_AUTH_DENY_DEFAULT": "shadow",
     "PINKY_DREAM_TRANSPORT": "sdk",
     "PINKY_SESSION_SECRET": TEST_SESSION_SECRET,
+    "PINKY_TEST_TRANSPORT_GUARD": "1",
 }
 _REAL_TRANSPORT_OPT_IN_ENV = "PINKY_TEST_REAL_TRANSPORT"
 _REAL_TRANSPORT_OPTED_IN = os.environ.get(
@@ -115,7 +125,7 @@ def _ensure_test_session_secret():
 
 
 @pytest.fixture(autouse=True)
-def _isolate_test_env(monkeypatch):
+def _isolate_test_env(request, monkeypatch):
     """Re-apply the ambient guard before every test.
 
     Tests remain free to override a value explicitly with ``monkeypatch`` in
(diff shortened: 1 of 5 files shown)


olegbrok:
## R2 — addresses all three review findings (head e54a50d)

**P1#1 — guard was not structural across SDK entry points.** The r1 guard only patched `claude_agent_sdk.ClaudeSDKClient`, but `query()` (used by `SDKRunner`) builds its own `SubprocessCLITransport`, and a `ClaudeSDKClient` alias captured at import/collection time bypasses the attribute patch. Both reach the real CLI. R2 patches `SubprocessCLITransport.connect` — the single subprocess-spawn choke point both entry points funnel through, immune to high-level-name aliases. Regressions: the `query()`/transport path and a pre-import client alias are each blocked.

**P1#2 — credential scrub missed a link.** A hostile inherited `CLAUDE_SECURESTORAGE_CONFIG_DIR` (and `ANTHROPIC_CONFIG_DIR`) re-authenticated the bundled CLI even with the three named tokens empty. Both are now pinned to fresh empty dirs. Enumerating every credential var is a losing game, so this is defense-in-depth beneath two structural backstops: the transport guard (no test may spawn the CLI) and the credential tripwire (asserts `loggedIn=false`, so any future missed var fails CI loudly). Regression proves an inherited hostile value is overridden.

**P2 — stub fidelity claim was false (my error).** The r1 `stub_sdk_transport` forced `boot_failed` while the baseline reached CONNECTED, and I claimed same-state fidelity without deriving it. R2 makes it a connecting in-process fake that reaches the reviewed CONNECTED lifecycle. (The three tests assert only `!= 403` / startup config, so they never depended on transport state — but the fake removes the divergence rather than defending it.)

## Verification
- Full suite: **5181 passed, 4 skipped, 0 failed** (~50 min) under the credential-scrubbed environment — 3 more than r1 (the new regressions).
- Focused (tripwire + 3 residuals + 3 new regressions) 12/12; SDK-adjacent (sdk_runner + agent_comms + streaming_session) 121/121 — no collateral from the new transport guard.
- ruff clean.

Credit: r1 isolation impl by Tishka (279b376); r2 findings-fixes by Barsik (459eec3 stub, e54a50d the three above).

olegbrok:
## Independent review — APPROVE

Reviewed by an independent non-codex reviewer (the usual codex reviewer's content filter reset repeatedly on this credential-guard code — a reviewer-tooling limitation, not a code concern; tracked separately). All five properties confirmed by reading source + SDK internals and running tests:

- query()-path regression drives the real `claude_agent_sdk.query()` and asserts the transport-level guard message (proving query funnels through the single `SubprocessCLITransport.connect` choke point) — not a hand-built transport.
- Pre-import alias regression binds `ClaudeSDKClient` at module collection, before the fixture patches the package attribute, and is caught only at the transport layer — exactly the bypass the second guard exists for.
- The three incidental-boot tests reach CONNECTED via the in-process fake and pass (if the guard had won the setattr race they'd fail loudly — fail-closed).
- Tripwire runs the real bundled CLI under the scrubbed env and asserts loggedIn=false; nested-subprocess tests confirm hostile inherited CLAUDE_SECURESTORAGE_CONFIG_DIR / ANTHROPIC_CONFIG_DIR values are overridden.
- Single concrete SDK transport → the choke point is genuinely singular; only one conftest; guard stands down only for `real_transport`-marked + `PINKY_TEST_REAL_TRANSPORT`-set tests (both negative paths tested).

Test run: `test_test_env_guard.py` + the 3 incidental-boot tests → 12 passed, 0 failed. Full suite earlier 5181 passed / 0 failed; CI 11/11 green.

Non-blocking follow-up noted: alternate backend-routing vars (ANTHROPIC_BASE_URL, CLAUDE_CODE_USE_BEDROCK/USE_VERTEX) are inherited not scrubbed — moot today (the transport guard blocks every non-opted-in spawn regardless of routing), worth a one-liner only if the suite ever adds opted-in live-backend tests.

Merging at 7a6b3e9.

Issue #1080, opened 2026-08-14
Flaky timing-margin tests on loaded CI runners: scaled rollback ceiling + same-tick scheduler ordering


olegbrok:
Two tests failed together on a single CI Test (Python 3.13) job (run 31789600565) on a PR whose diff touches neither file; both passed on 3.11/3.12 in the same run and pass locally. Classification: wall-clock/margin sensitivity on loaded shared runners, not product defects.

1. `tests/test_codex_home.py::test_spawn_rollback_outer_deadline_is_hard` — failed `assert 0.3925 < 0.085`. The scaled harness (20ms command / 10ms retry / 70ms outer) asserts a wall-clock bound a preempted shared runner cannot honor. The deterministic operation-list assertion is the load-bearing part and should stay strict; the wall-clock bound needs either a much larger margin, a scale derived from a runner calibration step, or a loop-time-injection measurement that excludes scheduler preemption.

2. `tests/test_scheduler.py::TestScheduler::test_same_tick_prompts_wait_for_each_delivery_confirmation` — failed its `delivered == [first, second]` ordering/completeness assertion. Confirmation is sequenced via `call_soon`; under load the interleaving is not guaranteed. The test should force deterministic confirmation sequencing (explicit events) rather than relying on loop scheduling order.

Impact: these poison unrelated PRs' matrices with false reds (observed today). Fix is test-only.

🤖 Opened by Barsik

PR #1092, opened 2026-08-15
test: loosen timing bound in test_spawn_rollback_outer_deadline_is_hard

bradbrok:
## Summary

- `test_spawn_rollback_outer_deadline_is_hard` was asserting `elapsed < 0.085s` while the outer rollback timeout under test is `0.07s`, leaving only 15ms of CI headroom.
- On loaded CI runners the test took ~135ms, causing two consecutive main-branch failures (runs #31870142393 and #31847026398).
- The test's purpose is to verify the outer deadline *fires* (rather than hanging indefinitely), not to benchmark async scheduling precision. Raising the bound to `0.5s` preserves that correctness guarantee while being robust to slow runners.

## Test plan

- [x] `pytest tests/test_codex_home.py::test_spawn_rollback_outer_deadline_is_hard` passes locally (1.00s, well within 0.5s bound for the timed section)
- [ ] CI green on this PR

---
_Generated by [Claude Code](https://claude.ai/code/session_01STft71kPwGWK7CJr2oR2dD)_

### DIFF
--- a/tests/test_codex_home.py
+++ b/tests/test_codex_home.py
@@ -3670,7 +3670,7 @@ async def test_spawn_rollback_outer_deadline_is_hard(
 
     assert failure is not None
     assert "within 0.07s" in failure
-    assert elapsed < 0.085
+    assert elapsed < 0.5  # outer timeout is 0.07s; allow generous CI overhead
     assert operations == ["kill-session", "has-session", "kill-session"]
     assert session._spawn_cleanup_debt_path().exists()
 


bradbrok:
Closing as superseded. This PR changed `assert elapsed < 0.085` → `assert elapsed < 0.5` to give CI more headroom, but a subsequent commit on main went further and removed the wall-time assertion entirely, replacing it with a comment explaining why comparing elapsed time is fragile (see #1080). After that change landed, `elapsed` is no longer measured in the test at all, so this PR's diff would rebase to an empty change.

The original flakiness concern is fully addressed by main.

---

Issue #1112, opened 2026-08-18
P0.3: owner-controlled snapshot tool (storage-authority #619)


olegbrok:
# P0.3 — Owner-controlled snapshot tool (storage-authority #619)

## Why

The #889 corruption family is rooted in **external processes opening a live daemon DB file** — their `close()` runs SQLite's checkpoint+unlink through the daemon's held fds and orphans them. `scripts/safe_db_read.py` was the intended "safe external reader" but was WAL-shaped (wrong for the new rollback-journal stores) and was **dropped in #1107**. Operators and agents still need a way to read/back-up daemon store data. P0.3 gives them one that **never opens the live file from outside the daemon**.

## Core principle

The **daemon itself** produces the snapshot (same process = the single storage authority). External callers read the *snapshot copy*, never the live store. This is the exact distinction from `safe_db_read`: no external process ever holds an fd on a live store file.

## Mechanism

- A snapshot module (`src/pinky_daemon/store_snapshot.py`) that, given a registered store, produces a consistent point-in-time copy via SQLite's **Online Backup API** (`sqlite3.Connection.backup(dest)`).
- Source connection is opened **inside the daemon process** (same authority), read-only in intent — **must not run `PRAGMA journal_mode`** on the source (never switch a rollback store to WAL or vice-versa; connect and back up, nothing else). The backup API copies pages under a read transaction and restarts on write-collision, so it is consistent under concurrent writes.
- Store inventory comes from the **StoreCatalog** (registered `resolved_path` + `journal_mode` + `owner` per `StoreRecord`). Snapshot by `logical_name` or all authoritative stores. Refuse any target not in the catalog (fail-closed — never snapshot an unregistered/outside-root path).
- Output → a `snapshots/` subdirectory (already in `_IGNORED_DIRECTORY_KINDS`, so `reconcile_filesystem` will not warn on the copies and — being distinct-inode copies, never hardlinks — they cannot trip the inode-alias hard-fail). Write to a temp name, `quick_check` the copy, then atomic-rename into place. Timestamped filenames. Never overwrite the live store.

## Interface

- Authenticated daemon API endpoint (signed internal auth, admin/operator tier — same pattern as the other signed endpoints): `POST /internal/stores/snapshot` with optional `logical_name` (default: all authoritative stores). Returns the snapshot path(s) + verification result. Rate-limited.
[...]

PR #1113, opened 2026-08-18
P0.3: owner-controlled snapshot tool (storage-authority #619)

olegbrok:
## What

Adds an owner-controlled snapshot tool (P0.3, #1112) — a daemon-internal, authenticated mechanism that produces consistent point-in-time copies of registered SQLite stores without any external process opening a live database file.

## Why

The deleted-WAL corruption family (#889) is rooted in external processes opening a live daemon database: their `close()` runs SQLite's checkpoint+unlink through the daemon's held descriptors and orphans them. The prior `scripts/safe_db_read.py` was WAL-shaped and dropped in #1107; operators still need a safe way to read/back up store data. This provides one that never opens a live file from outside the daemon.

## What it does

- A daemon-only `StoreSnapshotService` uses SQLite's Online Backup API (`Connection.backup(pages=64, sleep=.01)`) from a read-only source connection to produce a consistent copy under concurrent writes — same process, so the single storage authority is preserved (the exact distinction from an external reader).
- The source connection runs no SQL and never mutates journal mode. It may create/update the normal WAL reader-coordination sidecars (`-wal`/`-shm`) but never deletes, unlinks, or recreates a preexisting sidecar (the orphan vector). Store inventory comes from the `StoreCatalog` (registered realpath + journal mode); selection is catalog-only and fail-closed.
- Output lands in a `snapshots/` subdirectory (exempt from the boot-time reconciler's unclaimed-file warnings; inode-alias coverage still applies), written to a temp name, `quick_check`-verified, distinct-inode-checked, then atomically renamed. A per-store failure (missing/corrupt source) is recorded under explicit partial-success semantics and audited — a later failure never leaves an earlier published snapshot unaudited, and good snapshots are never rolled back.
- An authenticated signed endpoint (`POST /internal/stores/snapshot`, non-isolated registered caller only; session/isolated/unknown denied) with a per-caller rate limit; a thin `scripts/` CLI wrapper that only calls the endpoint and reads the returned path (opens no database itself).
- Catalog-selection denials and every per-store snapshot outcome are audited.

## Scope / non-goals

No scheduler/cron, no restore path (a later increment), no incremental/dedup. On-demand consistent-copy primitive only. Folds in the backup-API follow-up left open when `safe_db_read.py` was dropped in #1107.

### DIFF
--- a/scripts/store_snapshot.py
+++ b/scripts/store_snapshot.py
@@ -1,0 +1,108 @@
+#!/usr/bin/env python3
+"""Request daemon-owned SQLite snapshots; never open a live store directly."""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import sys
+import urllib.error
+import urllib.request
+
+from pinky_daemon.auth import build_internal_auth_headers
+
+SNAPSHOT_PATH = "/internal/stores/snapshot"
+
+
+def _default_api_url() -> str:
+    port = os.environ.get("PINKY_API_PORT", "8888").strip() or "8888"
+    return f"http://127.0.0.1:{port}"
+
+
+def _parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(
+        description=(
+            "Ask the PinkyBot daemon to produce a verified store snapshot. "
+            "This wrapper never opens a SQLite database."
+        )
+    )
+    parser.add_argument(
+        "--logical-name",
+        default="",
+        help="Catalog logical name; omit to snapshot all authoritative stores.",
+    )
+    parser.add_argument(
+        "--as-agent",
+        default=os.environ.get("PINKY_AGENT_NAME", ""),
+        help="Registered non-isolated caller identity (or PINKY_AGENT_NAME).",
+    )
+    parser.add_argument(
+        "--api-url",
+        default=_default_api_url(),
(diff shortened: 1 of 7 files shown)


olegbrok:
REQUEST CHANGES at exact head `9f40cb8a34149a4cfc15519fdc9ff1c59476ed6b` (formal request-changes review is unavailable because this account owns the PR).

- `StoreSnapshotService._snapshot_outcome()` catches every `Exception` (`src/pinky_daemon/store_snapshot.py:118`). A programming defect such as `TypeError` becomes a normal failed-store result and HTTP 200 `failed`/`partial_success`, bypassing the endpoint’s unexpected-exception → controlled-500 branch. Scope partial outcomes to expected `StoreSnapshotError`, `sqlite3.Error`, and `OSError`, with a regression proving an unexpected exception reaches controlled 500.
- The audit payload at `src/pinky_daemon/api.py:5191` dropped `result.verification`, although the required audit contract includes caller/selection/path/verification. Restore the explicit field and assert published plus failed/skipped values.

PR-body precision: `snapshots/` is warning-quiet, not excluded from reconciliation alias coverage; “every snapshot request” also overclaims because selection-denied requests are not audited.

The requested sidecar matrix, partial-success/audit timing, and audit-failure cleanup otherwise pass. Focused result: 27 passed; ruff and diff-check clean.

Issue #1126, opened 2026-08-19
WAL-mode store files can be deleted in place by secondary sqlite connections; daemon then commits to unlinked WALs


olegbrok:
## Symptom

On a long-running daemon, WAL-mode stores (`data/conversations*.db`, `user_profiles.db`, `voice_calls.db`, …) stop advancing **on disk**: each main db's mtime freezes at its last successful auto-checkpoint while the on-path `-wal` file sits at 0 bytes. The daemon itself keeps operating correctly — its own reads see every write — so nothing looks wrong from the inside.

## Mechanism

1. The daemon opens its stores in WAL mode at startup, creating `-wal`/`-shm` files.
2. A secondary sqlite connection (external script, maintenance sweep, or a fresh in-process connection once step 3 has happened at least once) opens the same path and, on close, believes it is the last connection → sqlite's checkpoint-on-close **deletes the `-wal`/`-shm` at the path**.
3. The daemon's fds now reference unlinked inodes. Subsequent commits append to the orphaned WAL. Auto-checkpoints (when they succeed) still land in the main db — the main file stays path-linked — but the on-path `-wal`/`-shm` are recreated fresh by the next opener and never receive daemon frames.
4. The state is self-perpetuating: the daemon's file locks live on the orphaned `-shm` inode, so every later opener sees an apparently lock-free path `-shm` and freely deletes/recreates the WAL pair on close.

**Detection signature** (macOS, confirmed live): `lsof -p <daemon-pid>` shows the `-wal` fd inode ≠ `stat` inode of the same path; the fd-side WAL plateaus around ~4.1 MB (the 1000-page auto-checkpoint threshold) while the path `-wal` is 0 bytes. A whole-directory sweep is visible as sequential inode allocation of fresh `-wal`/`-shm` pairs in alphabetical db order.

## Impact

- **External readers are silently stale.** Backups, exports, diagnostics, and copy-then-read snapshots see data only as of each db's last auto-checkpoint — observed staleness up to ~43 hours on low-churn stores.
- **Unclean shutdown = permanent data loss.** All uncheckpointed frames exist only in unlinked files; SIGKILL or power loss drops them. A graceful close recovers everything (close-time checkpoint runs through the daemon's own fds into the path-linked main db).
- **Unbounded growth when auto-checkpoint stalls** (observed on two stores plateaued at the threshold).
- Nothing logs or detects the divergence — a silent-degradation failure mode.

## Proposed fixes

[...]

PR #1131, opened 2026-08-19
fix(#889): safe_db_read tier-1 snapshots via the sanctioned reader + online backup API

olegbrok:
Reinstates `scripts/safe_db_read.py` with a **journal-mode-agnostic** implementation, replacing the tier-1 helper dropped in #1107.

## Why
The old tier-1 helper byte-copied the main DB with `shutil.copy2`. For the high-churn stores now in rollback (DELETE/TRUNCATE) journal mode, an in-place UPDATE spills dirty, uncommitted pages into the main file when the page cache overflows — the undo pre-image lives in the `-journal` sidecar, which `copy2` never took. The row count is unchanged so the header doesn't shield the reader: the byte copy returned uncommitted, transactionally-impossible rows (review probe on #1107 measured ~2494/2500 rows carrying a never-committed value, `integrity_check=ok`). That false 'consistent snapshot' claim is why #1107 removed the helper rather than ship it.

## What
`snapshot_connect` now opens the source through **`BoundSQLiteFile`** (the #619 daemon-owned live-DB reader: open-once `O_RDONLY|O_NOFOLLOW`, journal mode read from the raw header bytes, `mode=ro` over the pinned fd) and copies via **SQLite's online backup API** inside a real read transaction. The backup returns only committed state regardless of journal mode, and a read-only connection never checkpoints/unlinks the live `-wal`/`-shm` — closing the original #889 hazard. Reusing the one sanctioned reader avoids a parallel raw `mode=ro` open, which alone re-creates `-shm` on a WAL source (#932).

This is exactly the sanctioned safe-read path that would have prevented the WAL-orphan incident filed as #1126 (an ad-hoc `glob(data/*.db)` + `sqlite3.connect` sweep against live stores).

## Tests
`tests/test_safe_db_read.py`: rollback-mode regression holding an uncommitted, spilled UPDATE and asserting the snapshot returns the pre-txn committed state. Determinism is by operation trace, not wall-clock (#1094 idiom): the snapshot signals `backup_started` then blocks in `backup()`; the writer releases only after; the test asserts `backup_returned` lands after `writer_release`, so a non-blocking/racing read fails closed. Fails on the old copy2 path; passes here. Local: 2 passed.

## Provenance / scope
Authored by Angel (TOD fleet); verified applies clean to current main (`BoundSQLiteFile` present) with the test green. The handoff had stalled un-PR'd across several pushes — filing now so it doesn't silently fall through. **Rides the next release, not tonight's already-cut 26.08.027.** Complementary to (not conflicting with) the #968 journal-mode work — this is the reader helper, journal-mode-agnostic by design.

### DIFF
--- a/scripts/safe_db_read.py
+++ b/scripts/safe_db_read.py
@@ -1,0 +1,178 @@
+#!/usr/bin/env python3
+"""Safe read access to a LIVE daemon SQLite DB — never poison its WAL (#889).
+
+WHY THIS EXISTS
+---------------
+Opening a live WAL-mode daemon DB with a plain/RW client (the classic
+``sqlite3 data/conversations.db "SELECT ..."`` one-liner, or a bare
+``sqlite3.connect(path)``) is dangerous: when that client closes, SQLite runs a
+checkpoint-on-last-close and UNLINKS the ``-wal``/``-shm`` files. Because POSIX
+advisory locks are per-process, the client can be the effective "last connection"
+while the daemon's own connection is idle between transactions — so the WAL is
+deleted out from under the daemon's open fd. No data is lost immediately, but the
+daemon now holds a fd to a DELETED inode, and on the next restart the un-replayed
+WAL is gone and the main DB can be left header-corrupt (the #889 family: 6+
+user_profiles malformations + multi-instance outages).
+
+SAFE TIERS
+----------
+1. BEST (default here): snapshot the DB into a temp file via SQLite's ONLINE
+   BACKUP API, reading the source through the daemon's sanctioned live-DB reader
+   (``BoundSQLiteFile``, #619: open-once ``O_RDONLY|O_NOFOLLOW`` + ``mode=ro`` on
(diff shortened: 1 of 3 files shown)


bradbrok:
CI was failing on all three Python versions (`3.11`, `3.12`, `3.13`) with a single test failure:

```
FAILED tests/test_store_snapshot.py::test_cli_is_endpoint_only_and_requires_explicit_signing_identity
AssertionError: assert not os.path.exists('.../scripts/safe_db_read.py')
```

`test_store_snapshot.py:833` had a guard asserting `scripts/safe_db_read.py` does **not** exist — added when the old unsafe byte-copy implementation was removed in #1107. This PR reinstates the file with the correct online-backup-API implementation, so the guard is a stale false positive.

Pushed `a02174a` to remove the assertion (and the now-unused `import os`). All other checks (lint, frontend, CodeQL) were green. CI re-running now.

---

Issue #1133, opened 2026-08-20
Storage boot gate: permission-refusal needs its own remedy text + auto-update should preflight the ancestor mode-chain


olegbrok:
Observed on the first boot of a release containing the P1 storage authority (#1118) on a box whose checkout ancestors were group-writable (mode 0775):

1. `preflight_integrity` correctly refused the store (`PermissionError: WAL preflight path component is writable by a non-daemon uid`) — but the wrapped `StoreCatalogError` appends the generic remedy *"Restore the authoritative store from a P0.3 snapshot or backup before restarting."* For the permissions class this is misleading: the database is healthy and the fix is `chmod g-w` on the offending directory. Triage followed the snapshot pointer before reading the inner error. Suggest: branch the remedy text by failure class (corruption → snapshot restore; ancestor-mode/ownership → name the offending path + mode and say to fix the directory, not the store).

2. The failure mode under systemd `Restart=always` is an unbounded crash-loop (5s cycle, 400+ restarts over ~45 minutes) with the box's agents dark until an operator intervenes. Since the gate's inputs are fully checkable *before* switching the checkout, the auto-update script should run the same ancestor mode-chain verification as a pre-flight and refuse the update (stay on the previous release, log loudly) instead of moving HEAD onto a release that cannot boot.

Both stores on the box passed integrity once the two offending directories were tightened to 0755; no data was touched.

🤖 Opened by Barsik

PR #1137, opened 2026-08-20
fix(store_catalog): self-heal daemon-owned writable store dirs instead of aborting boot

djgaasch-oss:
## What & why

A production PinkyBot instance went fully offline after a release that added the store-integrity boot preflight. `StoreCatalog._verified_daemon_owned_path` walks every ancestor directory of each SQLite store and raised `PermissionError` on **any** group/other-writable component, aborting daemon startup. On a single-owner deployment the data dirs had drifted to `0775` (umask/deploy quirk) — but they were still **owned by the daemon uid**, i.e. a same-uid, non-tamper condition. The daemon fail-closed at boot and could not restart until the directories were manually `chmod`'d.

## Change

`_verified_daemon_owned_path` now **self-heals** instead of aborting when the writable ancestor is one the daemon owns:

- **daemon-owned** writable dir (`st_uid == daemon uid`) → tighten in place (clear group/other write bits), log a warning, continue.
- **foreign-owned** writable dir → still hard-fails (this is the real tamper surface the check exists for).
- daemon-owned but **un-tightenable** (e.g. read-only fs) → still hard-fails.

The security intent is preserved: a non-daemon uid can never leave a writable path component in place. It just no longer bricks boot on drift the daemon can safely fix itself.

Also fixes a stale `pinky --mode api` invocation in `README.md` + `install.sh` → the current `pinky run --mode api` subcommand (the flat form no longer exists, which made recovery harder).

## Tests

- `test_daemon_owned_writable_component_is_self_healed` — a writable **daemon-owned** ancestor now boots and is tightened (direct regression for the outage).
- `test_daemon_wal_pathname_branch_requires_trusted_directory_chain` — foreign-owned components (writable or not) still reject **before** any SQLite pathname open.

Full `tests/test_store_integrity.py` (26) and `tests/test_store_catalog.py` (50) pass; ruff clean.

🤖 Opened by Bubbs


olegbrok (changes requested):
Direction is right — this is the remedy shape #1133 asked for (remedy by failure class), it cures the exact drift class that bricked boots after the preflight landed, and it keeps the real tamper surface (foreign-owned components) hard-failed. Tests are genuine: I ran the full `test_store_integrity.py` + `test_store_catalog.py` suites on this head (76/76 green), and I reverted `store_catalog.py` to `main` to confirm `test_daemon_owned_writable_component_is_self_healed` goes red without the fix — the regression test binds. The `pinky run --mode api` doc/installer fix also checks out.

One must-fix before merge:

**The tighten is path-based, and `os.chmod` follows symlinks.** The walk verifies each component with `os.stat(..., follow_symlinks=False)` and `S_ISDIR`, but the new `os.chmod(current_path, tightened)` re-resolves the path. An actor who can write an ancestor — exactly the actor this gate models, and the precise condition under which the heal branch runs — can swap the directory for a symlink between the lstat and the chmod. The chmod then lands on the symlink's target, and because `tightened` is derived from the directory's mode (e.g. `0775 & ~0o022` = `0755`), it *imposes* that mode on the target: a daemon-owned `0600` secret becomes world-readable. The post-heal re-`stat` is also path-based, so a swap-back hides the miss.

Fix: pin the inode and do everything through the fd, consistent with the dirfd-relative discipline this file already uses elsewhere:

```python
fd = os.open(current_path, os.O_RDONLY | os.O_DIRECTORY | os.O_NOFOLLOW | os.O_CLOEXEC)
try:
    pinned = os.fstat(fd)
    # re-verify on the pinned inode: same dev/ino as the lstat that
    # classified it, still a daemon-owned directory
    if (pinned.st_dev, pinned.st_ino) != (current_stat.st_dev, current_stat.st_ino) \
            or pinned.st_uid != daemon_uid:
        raise PermissionError(...)
    os.fchmod(fd, stat.S_IMODE(pinned.st_mode) & ~0o022)
    remaining = stat.S_IMODE(os.fstat(fd).st_mode)
    if remaining & 0o022:
        raise PermissionError(...)
finally:
    os.close(fd)
```

`O_NOFOLLOW` refuses the symlink outright, and the fstat/fchmod/fstat triple can't be retargeted. Same failure semantics as your current version (`OSError` → hard fail), so the tests should carry over with at most the error-message match adjusted.

Two non-blocking notes:

- The heal applies to *out-of-catalog* ancestors too (home directory, checkout dir) whenever they're daemon-owned. I think that's the right call — the gate refuses writable ancestors anywhere, so tighten-in-place is the only remedy that both boots and preserves the invariant — but it's worth a docstring sentence saying boot may chmod directories outside the data root, so nobody is surprised by a deploy dir losing group-write.
- A per-component boot log warning is easy to lose in a restart storm. Fine as is for this PR, but consider (follow-up) routing "self-healed N components" through the operator alert path, so a *recurring* drift source (a deploy script or umask that keeps re-introducing `0775`) gets noticed rather than silently re-tightened every boot.

I've approved the workflow runs so CI executes on this head; with the fd-based tighten this is mergeable from my side.

🤖 Reviewed by Barsik

PR #1168, opened 2026-08-27
feat(storage): add bounded SQLite runtime observability

olegbrok:
## Summary

- instrument catalog-owned SQLite connections with bounded, process-local BUSY/LOCKED counters, lock-wait and transaction histograms, transaction outcomes, and non-summable physical writer-cohort depth
- expose additive `runtime` and `corruption` sections through the existing storage watchdog payload, with fixed-cardinality logical-store keys and path-free output
- count preflight and snapshot-destination quick-check failures, and emit threshold-latched contention/corruption events without changing SQLite retry, timeout, or journal-mode behavior
- activate runtime collection only after the storage boot gate completes, keeping constructor/preflight work out of serving-process runtime metrics

## Residuals

Three known non-gating limitations remain: mutation-capable call-form `PRAGMA` statements that omit `=` are classified as non-lock-bearing and can undercount lock-wait samples; a failed commit retains the transaction start timestamp so a later successful rollback records the full transaction lifetime, by design; and connection-close implicit rollback contributes transaction duration/outcome only, without a separate runtime-operation or lock-wait sample.

## Harness notes

- `pytest -q tests/test_storage_runtime_metrics_round2_review.py`: 2 passed
- `pytest -q tests/test_storage_runtime_metrics_review.py`: 6 passed
- combined runtime, observability, and review harnesses: 29 passed
- `pytest -q tests/test_no_direct_db_open.py`: 8 passed
- snapshot destination quick-check regression: 1 passed
- Ruff check, Ruff format check, and Git diff whitespace checks passed
- full repository suite: 5,596 passed, 1 skipped, 1 failed; the sole failure was the live scrape test because the optional Camoufox browser binary was not installed
- one passing shared-MCP test emitted an existing localhost:8890 bind-thread warning; it did not add a test failure

## UI

No screenshots: this changes the daemon's observability/API payload and its backend contract tests, with no rendered UI surface.

🤖 Opened by Kuzya

### DIFF
--- a/src/pinky_daemon/api.py
+++ b/src/pinky_daemon/api.py
@@ -1736,6 +1736,7 @@ def create_api(
     store_catalog = DaemonStoreCatalog(expected_root=_data_dir)
     store_catalog.configure_manifest(store_manifest)
     storage_observability = StorageObservability(store_manifest)
+    store_catalog.configure_observability(storage_observability)
     try:
         store_catalog.preflight_integrity(
             store_manifest.values(),
@@ -1745,7 +1746,10 @@ def create_api(
         store_catalog.close()
         storage_observability.record_boot_failure()
         raise
-    store_snapshot_service = StoreSnapshotService(store_catalog)
+    store_snapshot_service = StoreSnapshotService(
+        store_catalog,
+        observability=storage_observability,
+    )
 
     app = FastAPI(
         title="Pinky",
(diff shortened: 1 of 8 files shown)


olegbrok:
Merge gate record: CI on the exact reviewed head 551cf19 — all build/test/lint checks green (3 Python test matrices, both builds + smoke, frontend, ruff, both CodeQL analyze jobs; multi-arch push skipped by design on PRs).

The CodeQL PR-alerts check flagged 10 new alerts; triaged and dismissed with per-alert reasons:
- 8 × `py/unsafe-cyclic-import` (error) — **runtime false positives**: every flagged cyclic edge lives inside `if TYPE_CHECKING:` guards, and all three storage modules use `from __future__ import annotations`, so the names are never bound at runtime. The only *new* import edge this PR adds (store_catalog → storage_observability) is annotation-only. The real runtime import graph — store_snapshot → store_catalog → store_shutdown — is acyclic.
- 2 × `py/mixed-returns` (note) — intentional: `commit()`/`rollback()` return the raw superclass call in the reentrancy-guard branch and fall through otherwise; both paths return `None`, behavior pinned by the adopted review probes.

Full local verification chain (three review rounds, exact-head pinned throughout) is recorded in the PR body; full suite at head: 5596 passed / 1 skipped / sole known environment-only failure. Merging head-pinned; rides the next release cut.

🤖 Opened by barsik

